PR #3, opened 2020-09-18
minimal suite and test runner

lionelvillard:
- Add minimal suite and test runners
- TestContext includes injected client and informers
- Add the ability to invoke a subset of tests based on requirements levels (e.g --requirements-must=false)


pkg/framework/suite_runner.go
```diff
@@ -59,5 +75,29 @@ func (s *suite) Run() {
 		s.Configure(&BaseConfig{})
 	}
 
+	cfg = s.enableInjection()
+
 	os.Exit(s.m.Run())
 }
+
+func (s *suite) enableInjection() *rest.Config {
```
n3wscott, line 83: you can replace this with:

```
import 	"knative.dev/pkg/injection/sharedmain"

.. use:
	ctx := sharedmain.EnableInjectionOrDie(nil, nil)


```

lionelvillard: I tried but I'm getting `flag redefined: kubeconfig` and I don't know where it's coming from.

PR #5, opened 2020-09-21
work on components

lionelvillard:
- Bring in parts of Scott's rigging framework
- run `go test --build-images=true` to build and publish images. Default is false
- Add sequence stepper as an component example


pkg/components/sequencestepper/yaml.go
```diff
@@ -0,0 +1,45 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package sequencestepper
+
+const (
+	podTemplate = `
+apiVersion: v1
```
slinkydeveloper, line 21: Can we have the templates in separate files?

slinkydeveloper (commented):
The sequencestepper image seems exactly the same as the event transformer image in eventing

pkg/components/sequencestepper/yaml.go
lionelvillard: yes eventually. Can we do this in a different PR?

lionelvillard:
@slinkydeveloper sequencestepper was probably not the best first component to bring over. It's mostly used as an example to realize some of the ideas.

slinkydeveloper:
np @lionelvillard , just let's make sure that while working on the new test framework we don't bring in old bloat :smile:

PR #6, opened 2020-09-22
better configuration, add github component

lionelvillard:
- Load default configuration from file
- better configuration structure
- Require now takes a resource context and configuration for cluster-scoped component installation/verification
- Bring in Scott's rigging manifest package


pkg/components/eventing/sources/github/github.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"fmt"
+
+	"knative.dev/reconciler-test/pkg/config"
+	"knative.dev/reconciler-test/pkg/framework"
+)
+
+const (
+	artifactURITemplate = "https://github.com/knative/eventing-contrib/releases/download/v%s/github.yaml"
```
n3wscott, line 27: I think this is going the wrong direction. What is the plan here?

lionelvillard: the plan is to support installation from source, from a nightly build, or from a released artifact. For instance, we want to be able to check that the serving main branch is compatible with previously released GitHub source. Does this makes sense?

Also I'm pushing on the idea of having this framework covering all test phases starting from [phase 2](https://github.com/knative/eventing/issues/3903). It does not mean that when running the tests all phases are enabled. The idea is to first prepare the cluster (e.g. `go test ./... --setup-cluster=true`) and then run tests as a separate step. The benefit of this approach is to specify what you need as you write tests. No need to go modifying some obscure bash scripts or GitHub actions.

PR #8, opened 2020-09-24
conformance-lib poc

dprotaso:
Moving my knative.dev/pkg PoC to reconciler-test for review

/assign @n3wscott @lionelvillard @devguyio

### DIFF
--- a/pkg/test/config.go
+++ b/pkg/test/config.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package test
+
+import (
+	"flag"
+
+	"k8s.io/client-go/kubernetes"
+	"knative.dev/reconciler-test/pkg/test/environment"
+	"knative.dev/reconciler-test/pkg/test/feature"
+	"knative.dev/reconciler-test/pkg/test/requirement"
+)
+
+type Config interface {
+	Features() feature.States
+	Environment() environment.Settings
+	Requirements() requirement.Levels
+	KubeClient() kubernetes.Interface
+	ResetClients()
+}
+
+type BaseConfig struct {
+	requirement.Levels
+	environment.Settings
+	feature.States
+
+	kube kubernetes.Interface
+}
+
+func (c *BaseConfig) AddFlags(fs *flag.FlagSet) {
+	c.Levels.AddFlags(fs)
+	c.Settings.AddFlags(fs)
+	c.States.AddFlags(fs)
+}
+
+func (c *BaseConfig) Features() feature.States {
+	return c.States
+}
+
+func (c *BaseConfig) Environment() environment.Settings {
+	return c.Settings
+}
+
+func (c *BaseConfig) Requirements() requirement.Levels {
+	return c.Levels
+}
+
+func (c *BaseConfig) ResetClients() {
+	c.kube = kubernetes.NewForConfigOrDie(c.Environment().ClientConfig())
+}
+
+func (c *BaseConfig) KubeClient() kubernetes.Interface {
+	return c.kube
+}
--- a/pkg/test/context.go
+++ b/pkg/test/context.go
@@ -1,0 +1,94 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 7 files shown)


pkg/test/config.go
```diff
@@ -0,0 +1,68 @@
...
+	KubeClient() kubernetes.Interface
+	ResetClients()
+}
+
+type BaseConfig struct {
+	requirement.Levels
+	environment.Settings
+	feature.States
+
+	kube kubernetes.Interface
```
dprotaso, line 41: If we think tests are going to hit multiple clusters in Knative then we should move this out of the BaseConfig.

Maybe something like `SingleClusterConfig` - For now I haven't seen examples in Knative tests where we hit multiple clusters

pkg/test/config.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (c *BaseConfig) AddFlags(fs *flag.FlagSet) {
+	c.Levels.AddFlags(fs)
+	c.Settings.AddFlags(fs)
+	c.States.AddFlags(fs)
+}
```
dprotaso, line 48: Note: the goal of this is to move flags out of `init` and global variables in common packages. There's a bunch in `knative.dev/pkg/test` & `knative.dev/serving/test`.

This makes it hard to discover what can be configured for a test and I've seen setup code access flags directly - thus reaching around the test.

dprotaso: A second goal - allow downstream configs to easily add their own flags

pkg/test/config.go
```diff
@@ -0,0 +1,68 @@
...
+	"knative.dev/reconciler-test/pkg/test/feature"
+	"knative.dev/reconciler-test/pkg/test/requirement"
+)
+
+type Config interface {
+	Features() feature.States
+	Environment() environment.Settings
+	Requirements() requirement.Levels
+	KubeClient() kubernetes.Interface
+	ResetClients()
```
dprotaso, line 33: Since we run tests in parallel we don't want our k8s clients to hit client-side rate limiting.

pkg/test/config.go
```diff
@@ -0,0 +1,68 @@
...
+	"knative.dev/reconciler-test/pkg/test/environment"
+	"knative.dev/reconciler-test/pkg/test/feature"
+	"knative.dev/reconciler-test/pkg/test/requirement"
+)
+
+type Config interface {
+	Features() feature.States
+	Environment() environment.Settings
+	Requirements() requirement.Levels
+	KubeClient() kubernetes.Interface
```
dprotaso, line 32: Assumption here is these conformance/e2e test will do something with the k8s API

pkg/test/context.go
```diff
@@ -0,0 +1,94 @@
...
+	return t.invoke(feature.Stable, name, f)
+}
+
+func (t *T) Run(name string, f func(*T)) bool {
+	return t.invoke(feature.All, name, f)
+}
+
+func (t *T) ObjectNameForTest() string {
+	return helpers.ObjectNameForTest(t)
+}
```
dprotaso, line 79: there's a bunch of helpers from `knative.dev/pkg` test we can use

pkg/test/environment/environment.go
```diff
@@ -0,0 +1,128 @@
...
+	"os/user"
+	"path"
+	"text/template"
+	"time"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type Settings struct {
```
dprotaso, line 33: we could split out some of these properties

pkg/test/context.go
```diff
@@ -0,0 +1,94 @@
...
+	"knative.dev/reconciler-test/pkg/test/requirement"
+)
+
+func NewContext(t *testing.T, c Config) T {
+	c.ResetClients()
+	return T{Config: c, T: t}
+}
+
+type T struct {
+	Config
```
dprotaso, line 33: Generations would be nice here since we can qualify `Config`'s concrete type

dprotaso:
/hold

I'll stand up some serving & networking PRs tomorrow

pkg/test/config.go
lionelvillard: We are also trying to move away from typed clients because 1. they are very verbose and 2. they are not suitable for e2e testing. kubectl apply is better IMO. (e.g. `testcontext.Apply(FILENAME)`)

I don't think multi clusters is a requirement (at least for now)

pkg/test/config.go
lionelvillard: I've been using [this library](https://github.com/octago/sflags) to define and parse flags. The main reason is I want to be able to configure the test suites using a YAML file (because I'm expecting a lot of configuration parameters and hierarchies help) and to be able to override few parameters from the command line.

pkg/test/config.go
lionelvillard: yup I personally see this as an annoyance :-) But no strong feeling if it's only KubeClient.

pkg/test/config.go
dprotaso: Does `sflags` work with `go test` flags & arguments

lionelvillard: yes for arguments. I didn't try go test flags but _it should work_.

dprotaso: I guess to clarify - if you're hooking into golang's test lifecycle I think you're forced to use the same global flagset as the testing lib.

lionelvillard: Yup that's the case. It also means the YAML configuration file can also be used to set the testing flags, which is kind of cool. Example:

```yaml
test:
  timeout: "15m"
```

n3wscott:
ready to go?

dprotaso:
lol what timing

closing in favour of 
https://github.com/knative-sandbox/reconciler-test/pull/17

PR #13, opened 2020-09-28
move sample test suite to test/sample

lionelvillard:


lionelvillard:
ping. @slinkydeveloper

test/sample/suite_test.go
```diff
@@ -41,7 +41,6 @@ func TestMain(m *testing.M) {
 		NewSuite(m).
 		Configure(&myconfig).
 		Require(sequencestepper.Component).
-		Require(github.Component).
```
slinkydeveloper, line 44: Why you don't require github source anymore?

lionelvillard: for no good reason. I'll add it later on.

PR #14, opened 2020-09-28
Apply proposal

### LINKED ISSUE #11: CreateFromYaml/CreateFromURI/KoApply needs templating
I like the approach of using yamls, but they need templating features. For example, I could have the same broker config for two different brokers, but i need to change the name of the created resource. For that, we need some basic templating capabilities

lionelvillard:
Fixes #11

- Add Apply, ApplyOrError, Delete and DeleteOrError. 
- Each command above supports retrieving manifest from various sources (in-memory, string, and url) and supports templating


slinkydeveloper (approved):
I love it!

PR #15, opened 2020-10-02
add Get and GetOrError functions

lionelvillard:


pkg/framework/resourcecontext.go
```diff
@@ -79,12 +79,22 @@ type ResourceContext interface {
...
+	// DeleteOrError deletes the resources returned by provider.
 	// When data is set, it instantiates the manifest template
 	// before deleting it.
 	// Returns an error when DeleteOrError fails
 	DeleteOrError(provider manifest.Provider, data ...interface{}) error
 
+	// Get fetches the resource of the given name and populates object.
+	// Returns object so that Get can be used in Gomega assertions.
+	// The test is marked as Fail when Get fails
+	Get(resource string, name string, object runtime.Object) runtime.Object
```
slinkydeveloper, line 91: Should this accept gvr in order to be forward compatible? (GroupVersionKind)

pkg/installer/kubectl.go
```diff
@@ -0,0 +1,35 @@
...
+ */
+package installer
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Use ko to get a resource
+func KubectlGet(resource, name string, flags ...string) (string, error) {
```
slinkydeveloper, line 24: Can you use the pkg dynamicclient?

slinkydeveloper: `knative.dev/pkg/injection/clients/dynamicclient`

pkg/framework/resourcecontext.go
lionelvillard: I would like to keep the common case as simple as possible. Creating `gvr` object is cumbersome and I think it's better to keep the experience as closed as possible to kubectl, which does a lot of heavy lifting for us

pkg/installer/kubectl.go
lionelvillard: Maybe but it's more complicated and not sure it buys us anything. An alternative is to rely on kubectl go code to avoid creating a process. Anyhow we can always optimize later on if needed.

pkg/framework/resourcecontext.go
slinkydeveloper: Ok but my point is, what happens if we swap kubectl with dynamic client of something else?

lionelvillard: eventually we may want the something else to be this: https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/get/get.go#L448

slinkydeveloper: Still, my point is that we're introducing an API without a strong typing. I would prefer something like:

```golang
Get(resources.Group("messaging.knative.dev").Version("v1beta1").Kind("Channel"), "my-ch")
```

Not sure if this type has to be created in our framework or already exists somewhere, but IMO strong typing is a must here: it's verbose, but makes clear the API, helping the developer reasoning on the test. And it's forward compatible with whatever change we want to perform under the hood

Then we might have some alias for common ones, like `resources.Pod` or `resources.MessagingChannel`

Of course it's a personal opinion, feel free to ignore it.

lionelvillard: In this particular case, it's not about strong typing but about making sure the resource has the right format. In that respect, `resources.Group("messaging.knative.dev").Version("v1beta1").Kind("Channel")` is not better than `messaging.knative.dev.v1beta1.channels`. In both cases the test author can get it wrong (e.g. `resources.Group("messaging.v1beta.knative.ev").Version("latest").Kind("WhateverIWant")`)

`Get` should validate the resource has the right format (independently on how it is constructed).

What about getting this one in as is and get feedback from test authors?

slinkydeveloper: Sure we can have a chat at the test task force too

slinkydeveloper: Can you add this point in the task force meeting doc?

lionelvillard: done.

PR #17, opened 2020-10-06
conformance lib poc

dprotaso:
[Previously in #8](https://github.com/knative-sandbox/reconciler-test/pull/8/files#) I had `test.T`  separate from the `Config` interface. In practice this meant you had to cast the config to access certain properties.

I merged the two in this PR and changed the required signature when invoking subtests.

The functions that should be passed to subtests (`Run`, `Alpha`, `Must` etc..) should be take the same context that invoked the test

```go
// in my _test.go file
func RunConformance(t *testing.T) {
  ctx := MyTestContext{...}
  test.Init(ctx, t)
  ctx.Alpha("some alpha feature", TestAlphaFeature)
}

// somewhere else
func TestAlphaFeature(ctx *MyTestContext) {
  ctx.Must("do something", func(subctx *MyTestContext) {
     subctx.Fatal("failed to do something")
  })
}
```

I also realize eventing didn't use a ton of properties in the `environment.Settings` so I'm trying to break it up and not embed anything in the default `BaseContext`

### DIFF
--- a/pkg/test/context.go
+++ b/pkg/test/context.go
@@ -61,7 +61,9 @@ import (
 // To run your tests with T you can setup it as follows:
 //
 //   func TestConformance(t *testing.T) {
-//     myT := test.New(MyT{}, t)
+//     myT := MyT{}
+//     test.Init(MyT{}, t)
+//
 //     myT.Stable("some feature", TestFeature)
 //   }
 //
--- a/pkg/test/context.go
+++ b/pkg/test/context.go
@@ -26,21 +26,22 @@ import (
 	"knative.dev/reconciler-test/pkg/test/requirement"
 )
 
-// Context interface defines the methods required by a test context
-type Context interface {
-	// Copy should return a copy of the existing context
-	Copy() Context
+// C interface defines the methods required by T to clone and
+// setup subtests
+type C interface {
+	// Copy should return a copy of T
+	Copy() C
 
 	// Setup should initialize the context with a test
 	//
 	// Note: implementations shouldn't need to implement this
-	// if they embed a BaseContext struct
-	Setup(c Context, t *testing.T)
+	// if they embed a T struct
+	Setup(c C, t *testing.T)
 }
 
-// BaseContext extends testing.T with additional behaviour
+// T extends testing.T with additional behaviour
 // to test various requirement levels & feature states
-type BaseContext struct {
+type T struct {
 	*testing.T
 
 	RequirementLevels requirement.Levels
@@ -55,20 +56,20 @@ type BaseContext struct {
 	// ctx.Alpha("name", func(ctx *SomeTestContext) {...})
 	//
 	// This helps avoid excessive casting in downstream tests
-	c Context
+	c C
 }
 
-var _ Context = (*BaseContext)(nil)
+var _ C = (*T)(nil)
 
-// Setup implements the Context interface
-func (b *BaseContext) Setup(c Context, t *testing.T) {
-	b.c = c
-	b.T = t
+// Setup implements the C interface
+func (t *T) Setup(c C, test *testing.T) {
+	t.c = c
+	t.T = test
 }
 
-// Copy implements the Context interface
-func (b *BaseContext) Copy() Context {
-	cpy := *b
+// Copy implements the C interface
+func (t *T) Copy() C {
+	cpy := *t
 	return &cpy
 }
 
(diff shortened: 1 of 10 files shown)


pkg/test/context.go
```diff
@@ -0,0 +1,205 @@
...
+	"reflect"
+	"testing"
+
+	"knative.dev/pkg/test/helpers"
+	"knative.dev/reconciler-test/pkg/test/feature"
+	"knative.dev/reconciler-test/pkg/test/requirement"
+)
+
+// Context interface defines the methods required by a test context
+type Context interface {
```
n3wscott, line 30: can we call this something else. I want to use context.Context for client injection in e2e tests and I worry it will confuse the matter when talking about "context"

pkg/test/context.go
```diff
@@ -0,0 +1,205 @@
...
+		b.FeatureStates = feature.All
+	}
+
+	b.RequirementLevels.AddFlags(fs)
+	b.FeatureStates.AddFlags(fs)
+}
+
+// Must invokes f as a subtest if the context has the requirement level MUST
+func (b *BaseContext) Must(name string, f interface{}) bool {
+	b.Helper()
```
n3wscott, line 94: where does Helper come from?

dprotaso: embeded `testing.T`

this is here since if you pass a bad subtest function it'll fail and report the correct line

pkg/test/requirement/level.go
```diff
@@ -0,0 +1,145 @@
...
+	Should
+
+	// Should means that there may exist valid reasons in particular circumstances when the
+	// particular behavior is acceptable or even useful
+	ShouldNot
+
+	// May means that an item is truly optional
+	May
+
+	// TODO(dprotaso) Do we want these aliases?
```
n3wscott, line 46: I would say no.

pkg/test/context.go
dprotaso: I'll probably change it to `T` or something

dprotaso:
> GitHub Actions has encountered an internal error when running your job.
hmmm.. I can't seem to retrigger it in the UI

dprotaso:
nm found the button

pkg/test/requirement/level.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// Should means that there may exist valid reasons in particular circumstances when the
+	// particular behavior is acceptable or even useful
+	ShouldNot
+
+	// May means that an item is truly optional
+	May
+
+	// All flag enables all requirement levels
+	All = Must | MustNot | Should | ShouldNot | May
```
vaikas, line 47: I don't quite understand when you would use this flag?

pkg/test/requirement/level.go
```diff
@@ -0,0 +1,127 @@
...
+
+const (
+	// Must means that the definition is an absolute requirement of the specification.
+	Must Levels = 1 << iota
+
+	// MustNot means that the definition is an absolute prohibition of the specification.
+	MustNot
+
+	// Should means that there may exist valid reasons in particular circumstances to
+	// ignore a particular item
```
vaikas, line 36: nit: particular item => particular behavior for consistency with L40.

vaikas (commented):
Awesome sauce. tiny nits.

pkg/test/environment/cluster.go
```diff
@@ -0,0 +1,69 @@
...
+			defaultKubeconfig = path.Join(usr.HomeDir, ".kube/config")
+		}
+	}
+
+	// Allow for --kubeconfig on the cmd line to override the above logic
+	fs.StringVar(&s.KubeConfig, "env.cluster.kubeconfig", defaultKubeconfig,
+		"Provide the path to the `kubeconfig` file you'd like to use for these tests. The `current-context` will be used.")
+
+	fs.StringVar(&s.Namespace, "env.cluster.namespace", "",
+		"Provide the namespace you would like to use for these tests.")
```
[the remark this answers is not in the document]
dprotaso: I don't think this flag is used anywhere - it might have been in the past in serving

looks like it was but it's hardcoded now
https://github.com/knative/serving/commit/bfb3f5181b5ff18e4a16e9c158e8b9102404399c#diff-4a7296a57b6301815131c1b7506b3bc4L35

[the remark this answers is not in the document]
dprotaso: I'll drop it

pkg/test/requirement/level.go
dprotaso: I wasn't defaulting requirements ([which I am now](https://github.com/knative-sandbox/reconciler-test/pull/17/files#diff-1bda85529d562a2b1a5e5ed3504c42ecR82-R87)).

So if I wanted to enable all the tests via a flag you'd have to supply `-requirements.all`

pkg/test/requirement/level.go
dprotaso: lol I copied https://tools.ietf.org/html/rfc2119

dprotaso: i'll add a link to the RFC

pkg/test/requirement/level.go
dprotaso: it's a helper alias to run all tests

vaikas:
/lgtm

/hold
hold since others had comments.

pkg/test/context.go
```diff
@@ -0,0 +1,206 @@
...
+	"testing"
+
+	"knative.dev/pkg/test/helpers"
+	"knative.dev/reconciler-test/pkg/test/feature"
+	"knative.dev/reconciler-test/pkg/test/requirement"
+)
+
+// C interface defines the methods required by T to clone and
+// setup subtests
+type C interface {
```
dprotaso, line 31: I think this needs a better name - right now it looks like this downstream

```go
type ConformanceT struct {
	test.T

	Cluster environment.Cluster
	Images  environment.Images
	Ingress environment.Ingress
}

func (c *ConformanceT) Copy() test.C {
	return c
}

var _ test.C = (*ConformanceT)(nil)
```

dprotaso: 🤷‍♂️  

At a minimum I should include this example in the godoc

n3wscott: yeah, an example in the godoc will be nice

pkg/test/context_test.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+}
+
+func TestLevelInvocation(t *testing.T) {
+	cases := []struct {
+		name  string
+		level requirement.Levels
+		f     func(*mockContext, string, interface{}) bool
+	}{
+		{"Must", requirement.Must, (*mockContext).Must},
```
n3wscott, line 76: I don't see the negative type assertions where you test if a feature is alpha and you are running beta, or it is a should fails but passes.

dprotaso: > I don't see the negative type assertions where you test if a feature is alpha and you are running beta, or it is a should fails but passes.

Alpha features would be skipped in the scenario you describe - I can add these

dprotaso: should be good now

dprotaso:
Playing with this in https://github.com/knative/serving/pull/9742 leads to a few questions

### 1. Should test names be decorated with requirement levels & feature states

ie. right now it's
```
=== RUN   TestConformance/Service/multi_container
```

it could be
```
=== RUN TestConformance/Service/ALPHA_multi_container
```

It might make sense to do this only for non-stable & non-must/should requirements

```
=== RUN TestConformance/Service/ALPHA_multi_container/MAY_do_something
```

n3wscott:
/lgtm

Looks good from my side. Let's land it and tweak if needed.

PR #25, opened 2020-10-16
Change how we default feature states & requirement levels

dprotaso:
Defaulting to `All` in AddFlags didn't allow for a user to enable a specific level or state because internally we were ORing with a bitset


dprotaso:
missed a test
/hold

n3wscott (commented):
/lgtm
/approve

PR #26, opened 2020-10-19
Include a default parent context

dprotaso:
A lot of networking tests seem to use `context.Background()`.  Secondly this is a jumping off point for the context to be decorated in `Setup`

### DIFF
--- a/pkg/test/context.go
+++ b/pkg/test/context.go
@@ -17,6 +17,7 @@ limitations under the License.
 package test
 
 import (
+	"context"
 	"flag"
 	"reflect"
 	"testing"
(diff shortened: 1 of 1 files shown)


pkg/test/context.go
```diff
@@ -87,6 +88,10 @@ import (
 type T struct {
 	*testing.T
 
+	// If this context is not set it will be initalized to
```
tcnghia, line 91:
```suggestion
	// If this context is not set it will be initialized to
```

n3wscott (commented):
/lgtm
/approve

Issue #22, opened 2020-10-12
Port recordevents test image


slinkydeveloper:

slinkydeveloper:
Depends on https://github.com/knative/eventing/issues/4314

PR #30, opened 2020-10-23
Proposal: Knative E2E Test Framework

n3wscott:
Please see presentation: https://docs.google.com/presentation/d/1x-OJAQqAW7l5GDCO1WHWOIP7OMdbGXAYDN0uFtMGRlA/edit#slide=id.p


pkg/test/flags/flags.go
```diff
@@ -0,0 +1,63 @@
+// Splitting flag parsing to a self contained function returning a Config.
```
chizhg, line 1: Wrong header?

pkg/test/flags/flags_test.go
```diff
@@ -0,0 +1,105 @@
+// Unit tests for flag parsing.
```
chizhg, line 1: Ditto.

pkg/test/flags/flags.go
```diff
@@ -0,0 +1,63 @@
...
+// output of the flag.Parse is returned in output.
+// A special case is usage requests with -h or -help: then the error
+// flag.ErrHelp is returned and output will contain the usage message.
+func parseFlags(progname string, args []string) (config *Config, output string, err error) {
+	flags := flag.NewFlagSet(progname, flag.ContinueOnError)
+	var buf bytes.Buffer
+	flags.SetOutput(&buf)
+
+	var conf Config
+	flags.BoolVar(&conf.verbose, "verbose", false, "set verbosity")
```
chizhg, line 35: Could you explain a bit more on how these flags should be set and used? The current description isn't quite useful..

pkg/feature/name.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The Rigging Authors
+Copyright 2020 The Knative Authors
```
chizhg, line 2: The functions here seem to be a duplication of https://github.com/knative/pkg/blob/master/test/helpers/name.go, can we vendor instead?

pkg/test/flags/flags.go
n3wscott: the comments might be a bit off. I have not cleaned that part up yet

pkg/feature/name.go
n3wscott: We are going to pull all that out of pkg.

pkg/test/flags/flags.go
n3wscott: ah, this was code I did not intend to commit 😅 , it was an example I found of someone doing unit tests on flags.

n3wscott:
What do you think of renaming:

`Global Environment` -> `Environment`
`Environment` -> `Rig` 

Rig as in test rig... I still like that concept that the rig is the thing you are strapping your device under test into to perform the test...

slinkydeveloper:
@n3wscott nah, `GlobalEnvironment` and `Environment` are far more immediate IMO

tayarani:
@n3wscott I would also lean the `GlobalEnvironment` and `Environment` way. It's more clear what they represent, and having `Global` in the name does bring about extra caution when writing tests, i.e., it's well understood you should never be tampering with global state.

pkg/environment/magic.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func (mr *MagicEnvironment) Namespace() string {
+	return mr.namespace
+}
+
+func (mr *MagicEnvironment) Test(ctx context.Context, t *testing.T, f *feature.Feature) {
+	t.Helper() // Helper marks the calling function as a test helper function.
+	// Respect the flags, 0 means none.
+	//if mr.l == 0 {
```
tayarani, line 130: Should this be commented out? If it's intended as a guide, maybe move it to somewhere more appropriate?

pkg/environment/magic.go
```diff
@@ -0,0 +1,188 @@
...
+	//	mr.l = feature.All
+	//}
+	//if mr.s == 0 {
+	//	mr.s = feature.Any
+	//}
+
+	for _, timing := range feature.Timings() {
```
tayarani, line 137: This method is 2 nested loops and 2 local functions. Would it be better to refactor into multiple functions?

pkg/environment/namespace.go
```diff
@@ -0,0 +1,84 @@
...
+	interval = 1 * time.Second
+	timeout  = 45 * time.Second
+)
+
+// CreateNamespaceIfNeeded creates a new namespace if it does not exist.
+func (me *MagicEnvironment) CreateNamespaceIfNeeded() error {
+	c := kubeclient.Get(me.c)
+	nsSpec, err := c.CoreV1().Namespaces().Get(context.Background(), me.namespace, metav1.GetOptions{})
+
+	if err != nil && apierrors.IsNotFound(err) {
```
tayarani, line 41: Would it be better for readability if this was changed to `if err == nil || !apierrors.IsNotFound(err) { return nil }`. Then the rest of the method can be un-indented by one level.

tayarani: Also, I'm not sure the current logic is correct. If there is an error which is not the not found error, we should return err, not nil. The current code returns nil.

pkg/observer/recorder-collector/collector.go
```diff
@@ -51,15 +52,17 @@ type collector struct {
...
 	} else {
-		lister = c.client.CoreV1().Events(from.Namespace) // TODO: I do not understand how to do cluster scoped objects.
+		lister = c.client.CoreV1().Events(from.Namespace)
 	}
 	events, err := lister.List(ctx, metav1.ListOptions{})
 	if err != nil {
 		return nil, err
 	}
 
+	fmt.Println("Listed ", len(events.Items), " k8s events.")
```
tayarani, line 64: Was this left over from debugging?

test/example/main_test.go
```diff
@@ -30,38 +30,46 @@ import (
...
+
+	global = environment.NewGlobalEnvironment(ctx)
+
 	os.Exit(m.Run())
 }
 
 // This test is more for debugging the ko publish process.
 func TestKoPublish(t *testing.T) {
-	ic, err := installer.ProduceImages()
+	fmt.Println("TestKoPublish")
```
tayarani, line 55: Remove?

tayarani:
Looks pretty good. I put some minor comments.

n3wscott:
thanks for the review @tayarani !! I will fix this up and maybe we can land this first version.

pkg/environment/flags.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"knative.dev/reconciler-test/pkg/feature"
+)
+
+var (
+	s = new(feature.States)
+	l = new(feature.Levels)
+)
+
+func InitFlags(fs *flag.FlagSet) {
```
slinkydeveloper, line 33: I wonder, more than flags (or complementary to flags), can we use some yaml config files? We have so many stuff we're testing and flags are becoming really messy and hard to read https://github.com/knative/eventing/blob/master/test/e2e-conformance-tests.sh#L36.

Even env variables looks better, we could have a script to source to load them before starting `go test`

pkg/feature/feature.go
```diff
@@ -0,0 +1,153 @@
...
+	"fmt"
+	"testing"
+)
+
+type Feature struct {
+	Name  string
+	Steps []Step
+}
+
+type StepFn func(ctx context.Context, t *testing.T)
```
slinkydeveloper, line 30: Can step fn panic? is the panic recovered and recorded as failure?

pkg/feature/level.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	return b.String()
+}
+
+var LevelMapping = map[Levels]string{
+	Must:      "MUST",
+	MustNot:   "MUST NOT",
+	Should:    "SHOULD",
+	ShouldNot: "SHOULD NOT",
```
slinkydeveloper, line 78: From a test framework perspective, the `* NOT` are not levels, NOT is part of the predicate they're expressing.

pkg/feature/feature.go
```diff
@@ -0,0 +1,153 @@
...
+		T:    Assert,
+		Fn:   fn,
+	})
+}
+
+type Assertable interface {
+	Must(name string, fn StepFn) Assertable
+	Should(name string, fn StepFn) Assertable
+	May(name string, fn StepFn) Assertable
+	MustNot(name string, fn StepFn) Assertable
```
slinkydeveloper, line 100: Same comment as below, i expect `MustNot` and `ShouldNot` to be "wrappers" that negate the result of `stepFn`

pkg/observer/recorder-collector/collector.go
slinkydeveloper: You can keep it there, i'm gonna erase this stuff with #22

pkg/manifest/installer.go
```diff
@@ -0,0 +1,66 @@
...
+	"log"
+	"os"
+	"path"
+	"runtime"
+
+	"knative.dev/pkg/injection/clients/dynamicclient"
+	"knative.dev/reconciler-test/pkg/environment"
+)
+
+func InstallLocalYaml(ctx context.Context, base map[string]interface{}) (Manifest, error) {
```
slinkydeveloper, line 30: Looking at the example, should `Install` wait (or at least, be configurable to do so) for a resource to be completely up (like, for a pod, wait it to be up and running)

test/example/config/echo/echo.go
```diff
@@ -14,10 +14,35 @@ See the License for the specific language governing permissions and
...
+	)
+}
+
+// Output is the base output we can expect from a echo job.
 type Output struct {
 	Success bool   `json:"success"`
 	Message string `json:"msg"`
 }
+
+func Install(message string) feature.StepFn {
```
slinkydeveloper, line 40: I know this is a job, but the semantic of `Install`, throughout our components, should wait for the pod to be up and running?

test/example/config/producer/producer.go
```diff
@@ -0,0 +1,47 @@
...
+	"knative.dev/reconciler-test/pkg/manifest"
+)
+
+func init() {
+	environment.RegisterPackage(
+		"knative.dev/reconciler-test/test/example/cmd/producer",
+	)
+}
+
+func Install(sendCount int, svcName string) feature.StepFn {
```
slinkydeveloper, line 36: This one, should it wait for the deployment to be up?

test/example/config/recorder/recorder.go
```diff
@@ -0,0 +1,85 @@
...
+	recorder_collector "knative.dev/reconciler-test/pkg/observer/recorder-collector"
+)
+
+func init() {
+	environment.RegisterPackage(
+		"knative.dev/reconciler-test/images/recorder",
+	)
+}
+
+func Install(recorderName string) feature.StepFn {
```
slinkydeveloper, line 39: FYI I think there is a problem with this `StepFn`: It needs to return a context: When you install recordevents, you want to start the event recorder and keep it inside the context, so you can lately retrieve it. How can I do it here?
In any case, let's tackle this while solving the #22

slinkydeveloper (commented):
I think as a first pass this is interesting. Most of my comments are just ideas and doubts that can be solved in followups.

pkg/environment/magic.go
n3wscott: It was an issue I found with the flag parsing, thanks.

pkg/environment/magic.go
n3wscott: Ack, and let's do a followup.

pkg/environment/namespace.go
n3wscott: Updated.

pkg/environment/flags.go
n3wscott: I think we can look at that. but as a followup.

pkg/feature/level.go
n3wscott: yes I know. @dprotaso  insisted it follow the spec language.

pkg/manifest/installer.go
n3wscott: My thinking is if you need sequenced steps like that you break it into a feature and run more than one feature on an environment

test/example/config/echo/echo.go
n3wscott: Not always, because you might be wanting to test that things are late-bindable.

test/example/config/producer/producer.go
n3wscott: See how I use this in the example. it is a feature and one phase is install, then it waits for it to be ready. The following features use it and then assert that it captured stuff

test/example/config/recorder/recorder.go
n3wscott: I was assuming you do this through the environment, not the context.

pkg/environment/images.go
```diff
@@ -30,6 +30,7 @@ var packaged sync.Once
 
 // RegisterPackage registers an interest in producing an image based on the
 // provide package.
+// Can be called multiple times with the same package
```
vaikas, line 33: Is the idea here that you RegisterPackage 1+ more times, then you call the ProduceImages? Maybe an example if that's the case?

pkg/environment/images.go
```diff
@@ -45,6 +46,14 @@ func RegisterPackage(pack ...string) {
 	}
 }
 
+// WithImages will bypass ProduceImages() and use the provided image set instead.
+// Should be called before ProduceImages(), if used, likely in an init() method.
```
vaikas, line 50: Could you document what the images is, a pointer to something like gcr.io/... or something else?

pkg/environment/interfaces.go
```diff
@@ -0,0 +1,72 @@
...
+	// can be listed in env.References() or be cleaned up in env.Finish().
+	// This can be one way a feature communicates with future features run in
+	// the same environment.
+	Reference(ref ...corev1.ObjectReference)
+
+	// References returns the list of known object references that have been
+	// installed in the environment.
+	References() []corev1.ObjectReference
+
+	// Finish signals to the enviornment no future features will be run. The
```
vaikas, line 68: nit: environment

pkg/environment/interfaces.go
```diff
@@ -0,0 +1,72 @@
...
+	FeatureState() feature.States
+
+	// Images returns back the name to container image mapping to be used with
+	// yaml template parsing.
+	Images() map[string]string
+	// TemplateConfig returns the base template config to use when processing
+	// yaml templates.
+	TemplateConfig(base map[string]interface{}) map[string]interface{}
+
+	// Reference registers a object reference to the environment, so that it
```
vaikas, line 58: nit: an object reference

pkg/environment/magic.go
```diff
@@ -0,0 +1,187 @@
...
+	s feature.States
+
+	images           map[string]string
+	namespace        string
+	namespaceCreated bool
+	refs             []corev1.ObjectReference
+}
+
+func (mr *MagicEnvironment) Reference(ref ...corev1.ObjectReference) {
+	if mr.refs == nil {
```
vaikas, line 59: I think you can skip this and just do:
mr.refs = append(mr.refs, ref...)

append does the right thing.

pkg/environment/magic.go
```diff
@@ -0,0 +1,187 @@
...
+		mr.refs = make([]corev1.ObjectReference, 0)
+	}
+	mr.refs = append(mr.refs, ref...)
+}
+
+func (mr *MagicEnvironment) References() []corev1.ObjectReference {
+	if mr.refs == nil {
```
vaikas, line 66: I think you don't need this either since the array nil value should do the right things.

pkg/feature/name.go
```diff
@@ -73,6 +73,11 @@ func RandomString() string {
 	return string(suffix)
 }
 
+// MakeRandomK8sName
+func MakeRandomK8sName(prefix string) string {
```
vaikas, line 77: I assume that somewhere truncation happens, just jotting this down. Just want to make sure we don't end up with too long of a name.

pkg/k8s/logs.go
```diff
@@ -0,0 +1,57 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes"
+)
+
+func LogsFor(client kubernetes.Interface, namespace, name string, gvr schema.GroupVersionResource) (string, error) {
+	// Get all pods in this namespace.
+	pods, err := client.CoreV1().Pods(namespace).List(context.Background(), metav1.ListOptions{})
```
vaikas, line 32: Just curious why we list the pods and then find the pod again. Does GetLogs not handle if pod is not there? If it does, why spin through all of them first? Maybe comment if there's some subtlety here.

pkg/k8s/logs.go
```diff
@@ -0,0 +1,57 @@
...
+			if l, err := client.CoreV1().Pods(namespace).GetLogs(pod.Name, &corev1.PodLogOptions{}).DoRaw(context.Background()); err != nil {
+				logs = append(logs, err.Error())
+			} else {
+				logs = append(logs, string(l))
+			}
+		}
+	}
+
+	// Did we find a match like the given name?
+	if len(logs) == 0 {
```
vaikas, line 52: does GetLogs not return NotFound?

pkg/environment/flags.go
slinkydeveloper: Log it in an issue?

pkg/feature/level.go
slinkydeveloper: Don't get me wrong, but I think this is a mistake, It's like saying "We need a method for SHOULDAND for a spec sentence like SHOULD predicate_a AND predicate_b", you're mixing requirement levels with predicate operators, which are 2 very separate concepts in a test framework. Our test framework should provide predicate operators separately from requirement levels

pkg/manifest/installer.go
slinkydeveloper: :+1:

test/example/config/recorder/recorder.go
slinkydeveloper: Ok I think i miss your point here.

pkg/environment/interfaces.go
```diff
@@ -16,13 +14,33 @@ See the License for the specific language governing permissions and
...
+
+	Namespace() string
+
+	RequirementLevel() feature.Levels
+	FeatureState() feature.States
+
+	Images() map[string]string
+	TemplateConfig(base map[string]interface{}) map[string]interface{}
+
+	Reference(ref ...corev1.ObjectReference)
```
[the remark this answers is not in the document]
slinkydeveloper: I see, it's what in eventing we call Tracker today

test/example/config/recorder/recorder.go
slinkydeveloper: Didn't you sad in previous discussions that objects like `eventTracker` should live inside `ctx` and the user unwraps them from the context and performs assertions?

pkg/environment/flags.go
lionelvillard: +1 for yaml config files.

lionelvillard: this is a good simple library for handling flags: https://github.com/octago/sflags

pkg/environment/images.go
```diff
@@ -45,6 +46,14 @@ func RegisterPackage(pack ...string) {
...
+func WithImages(images map[string]string) {
+	packaged.Do(func() {
+		packageToImageConfig = images
+	})
+}
+
 // ProduceImages returns back the packages that have been added.
 // Will produce images once, can be called many times.
 func ProduceImages() (map[string]string, error) {
```
lionelvillard, line 59: We need to make sure this is optional since it takes time to produce images.

test/example/echo_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"github.com/google/uuid"
+	kubeclient "knative.dev/pkg/client/injection/kube/client"
+	"knative.dev/reconciler-test/pkg/environment"
+	"knative.dev/reconciler-test/pkg/feature"
+	"knative.dev/reconciler-test/pkg/k8s"
+	"knative.dev/reconciler-test/test/example/config/echo"
+)
+
+func EchoFeature() *feature.Feature {
```
lionelvillard, line 34: I feel like yet another Ginkgo framework to me. Sure the syntax might look nicer to some but by reading the code below it's not clear what's really happening.

test/example/echo_test.go
```diff
@@ -0,0 +1,86 @@
...
+	msg := fmt.Sprintf("hello %s", uuid.New())
+
+	f := new(feature.Feature)
+
+	f.Setup("install echo", echo.Install(msg))
+
+	f.Requirement("echo job is finished", func(ctx context.Context, t *testing.T) {
```
lionelvillard, line 41: Does Requirement execute the lambda right away or is it deferred?

test/example/echo_test.go
```diff
@@ -0,0 +1,86 @@
...
+		env := environment.FromContext(ctx)
+		client := kubeclient.Get(ctx)
+
+		if err := k8s.WaitUntilJobDone(client, env.Namespace(), "echo", time.Second, 30*time.Second); err != nil {
+			t.Errorf("failed to wait for job to finish, %s", err)
+		}
+	})
+
+	f.Alpha("pull logs off a pod").
```
lionelvillard, line 50: same here. Anonymous functions make it hard to understand when they are going to be called.

pkg/environment/flags.go
n3wscott: https://github.com/knative-sandbox/reconciler-test/issues/31

pkg/feature/feature.go
n3wscott: https://github.com/knative-sandbox/reconciler-test/issues/32

n3wscott: Maybe, I think we fix in a followup.

pkg/feature/level.go
n3wscott: Except the AND is not a spec word. It would be "SHOULD predicate_a and SHOULD predicate_b"

pkg/feature/name.go
n3wscott: https://github.com/knative-sandbox/reconciler-test/issues/33

pkg/k8s/logs.go
n3wscott: This was code copied out of pkg... idk

pkg/environment/images.go
n3wscott: it is optional, you can manually set the images.

test/example/echo_test.go
n3wscott: it is run when the step in the feature is tested in an environment.

test/example/echo_test.go
n3wscott: I was more showing it is possible, I don't think you are asking when it is run, but what the scope the function runs. The scope will be the normal rules go uses when functions are invoked.

pkg/feature/level.go
slinkydeveloper: I would love to hear other people opinion about that, @lionelvillard @tayarani ?

lionelvillard: +1 to follow what the rfc spec says. In this context `NOT` is not a predicate operator, it's just part of the phrase.

pkg/environment/interfaces.go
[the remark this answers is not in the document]
n3wscott: yes, though maybe it is something we might be able to infer later by looking at the namespace

test/example/config/recorder/recorder.go
n3wscott: I worry about ordering if you are going to assume the step can mutate the context

pkg/environment/images.go
n3wscott: The thinking here is you can include the two different deployments that have a dependency on a common image but are functionally different (like the event observer code that now can send or recieve) two packages might register that same image and use it in two different ways, and I was trying to avoid a `flags` redef issue.

pkg/environment/images.go
n3wscott: added some more words, but I think this needs to change a little.

n3wscott: https://github.com/knative-sandbox/reconciler-test/issues/34

pkg/k8s/logs.go
n3wscott: I think that bit of code might need to get plucked out of here, it might be more scope creep for the framework?

pkg/feature/feature.go
```diff
@@ -0,0 +1,153 @@
...
+		Name: fmt.Sprintf("%s %s", a.name, name),
+		S:    a.s,
+		L:    l,
+		T:    Assert,
+		Fn:   fn,
+	})
+}
+
+type Assertable interface {
+	Must(name string, fn StepFn) Assertable
```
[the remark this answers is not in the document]
n3wscott: I think we will be able to write that test and not have it be cumbersome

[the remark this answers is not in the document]
dprotaso: This seems dismissive and hand waving the problem away - as is it will be cumbersome 

Every assertion will need to pull info off the context - and some of these assertions are one off and there's no re-use

vaikas:
/lgtm
/hold
since others have comments, remove if you're fine...

Issue #34, opened 2020-11-12
Images produced should be provided to the templates with a bit more complexity to avoid shadowing


n3wscott:
Could you document what the images is, a pointer to something like gcr.io/... or something else?

_Originally posted by @vaikas in https://github.com/knative-sandbox/reconciler-test/pull/30#discussion_r520941041_

PR #35, opened 2020-11-12
Use ko:// to signal that we would like the test framework to produce images.

n3wscott:
tl;dr: `image: { .images.magicName }` --> `image: ko://some/go/package`

Fixes: https://github.com/knative-sandbox/reconciler-test/issues/34

This change takes some opinions out of the testing framework, we will use `ko://` to signal we want to make a container. 

This is overwritten if you use something like the following:

```
// The following could be used to bypass the image generation process.

import "knative.dev/reconciler-test/pkg/environment"

func init() {
	environment.WithImages(map[string]string{
		"ko://knative.dev/reconciler-test/images/recorder": "example.azurecr.io/kn/recorder@sha256:1938a18961f21262278dc03d18be190c630bef95a9ca0c62ca08af2528dbfd86",
		"ko://knative.dev/reconciler-test/test/example/cmd/echo": "example.azurecr.io/kn/echo@sha256:0933ea6bf2799b08653c0e729d63372629b42c69331125d2428edeb25c57637b",
		"ko://knative.dev/reconciler-test/test/example/cmd/producer": "example.azurecr.io/kn/producer@sha256:7ed12ab13b4828ab1a3797c52aea09c40681d6c79237e86ca9050041fd4ab008",

	})
}
```

At the moment, if WithImage is set, then ProduceImages will do nothing. We could do a followup to be more smart about this, but for now, let's try this out.

### DIFF
--- a/pkg/environment/images.go
+++ b/pkg/environment/images.go
@@ -69,9 +69,7 @@ func ProduceImages() (map[string]string, error) {
 				propErr = err
 				return
 			}
-			i := strings.Split(pack, "/")
-			// TODO: the current key is too small, we will need to follow-up and change the image to be more complex to avoid shadowing.
-			packageToImageConfig[i[len(i)-1]] = strings.TrimSpace(image)
+			packageToImageConfig["ko://"+pack] = strings.TrimSpace(image)
 		}
 	})
 	return packageToImageConfig, propErr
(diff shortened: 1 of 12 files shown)


n3wscott:
Ok, now works. I had overwritten the wrong file on string replace.

slinkydeveloper:
Sounds reasonable to me cc @lionelvillard 

/lgtm
/hold for another pass

PR #46, opened 2020-11-18
adding dependency method

n3wscott:
# Changes

There are times where we need to apply or validate a dependency to an environment but it is awkward to make that a `Test()` if it is intended to be pure setup, like making a namespaced component that will be used by some underlying infrastructure code (like rabbitmq brokers in a namespace, which are used by the Broker CO, but not directly by the test)

- :gift: Adding `env.Dependency()`.

/kind api-change

**Release Note**

```release-note
Adding `env.Dependency()`, just like `env.Test()` except we can make it clear it is not a Test, but some required other thing to be applied to the environment in the logs and results.
```

### DIFF
--- a/pkg/environment/interfaces.go
+++ b/pkg/environment/interfaces.go
@@ -37,6 +37,10 @@ type GlobalEnvironment interface {
 
 // Environment is the ephemeral testing environment to test features.
 type Environment interface {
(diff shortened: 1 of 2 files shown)


vaikas:
That looks reasonable, let's see how it looks in practice and revisit if things need to change?
/lgtm
/approve

PR #52, opened 2020-11-20
Running constraints of different timings

slinkydeveloper:

# Changes

- :broom: Now `Setup`, `Teardown` and `Requirement` run ordered and sequentially, while `Assert` is tagged to run in parallel. This makes the setup, teardown and requirement phases of the test predictable
- :gift: Test for the runtime constraints of the tests

### DIFF
--- a/pkg/feature/collapse.go
+++ b/pkg/feature/collapse.go
@@ -1,0 +1,76 @@
+package feature
+
+import (
+	"context"
+	"testing"
+)
+
+// CollapseSteps is used by the framework to impose the execution constraints of the different steps
+// Steps with Setup, Requirement and Teardown timings are run sequentially in order
+// Steps with Assert timing are run in parallel
+func CollapseSteps(steps []Step) []Step {
+	var setup *Step
+	var requirement *Step
+	var asserts []Step
+	var teardown *Step
+
+	for i, s := range steps {
+		if s.T == Setup {
+			setup = composeStep(setup, &steps[i])
+		} else if s.T == Requirement {
+			requirement = composeStep(requirement, &steps[i])
+		} else if s.T == Assert {
+			asserts = append(asserts, parallelizeStep(s))
+		} else if s.T == Teardown {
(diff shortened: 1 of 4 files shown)


pkg/feature/collapse.go
```diff
@@ -0,0 +1,76 @@
...
+// Steps with Assert timing are run in parallel
+func CollapseSteps(steps []Step) []Step {
+	var setup *Step
+	var requirement *Step
+	var asserts []Step
+	var teardown *Step
+
+	for i, s := range steps {
+		if s.T == Setup {
+			setup = composeStep(setup, &steps[i])
```
n3wscott, line 19: why not the way it was?

I think this is much more confusing in the output of the test:

```
=== test1 AND test2 AND test3
```

if that fails, how are you going to debug which is having the problem?

slinkydeveloper: That's fair, but I think we can solve it somehow in a followup, for example we could manually print these:

```
==== test1
==== test2
==== test3
```

slinkydeveloper: The important thing here is that we force the ordered and sequential execution

pkg/feature/collapse.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// CollapseSteps is used by the framework to impose the execution constraints of the different steps
+// Steps with Setup, Requirement and Teardown timings are run sequentially in order
+// Steps with Assert timing are run in parallel
+func CollapseSteps(steps []Step) []Step {
+	var setup *Step
+	var requirement *Step
+	var asserts []Step
+	var teardown *Step
```
vagababov, line 15:
```suggestion
	var (
	     setup, requirement, teardown *Step
	     asserts []Step
	)
```

pkg/feature/collapse.go
```diff
@@ -0,0 +1,76 @@
...
+
+	for i, s := range steps {
+		if s.T == Setup {
```
vagababov, line 18: switch here

PR #57, opened 2020-11-23
adding IsReady for downstream usage

### LINKED ISSUE #53: Create a KResource version of IsReady
This one can be abstracted and simplified, we could have in reconciler-test this:

```golang
func IsReady(gvr) func (name string, interval, timeout time.Duration) feature.StepFn {
return func (name string, interval, timeout time.Duration) feature.StepFn {
	return func(ctx context.Context, t *testing.T) {
		env := environment.FromContext(ctx)
		if err := k8s.WaitForResourceReady(ctx, env.Namespace(), name, gvr, interval, timeout); err != nil {
			t.Error("%v did not become ready, ", grv, err)
		}
	}
}
}
```

_Originally posted by @slinkydeveloper in https://github.com/knative/eventing/pull/4564#discussion_r528839570_

n3wscott:
# Changes

- Adding IsReady for downstream usage.

/kind enhancement

Fixes #53 

**Release Note**

```release-note
k8s.IsReady() is now usable.
```

### DIFF
--- a/pkg/k8s/steps.go
+++ b/pkg/k8s/steps.go
@@ -1,0 +1,38 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package k8s
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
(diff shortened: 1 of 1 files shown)


pkg/k8s/steps.go
```diff
@@ -0,0 +1,38 @@
...
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"knative.dev/reconciler-test/pkg/environment"
+	"knative.dev/reconciler-test/pkg/feature"
+)
+
+// IsReady returns a reusable feature.StepFn to assert if a resource is ready.
+func IsReady(gvr schema.GroupVersionResource, name string, interval, timeout time.Duration) feature.StepFn {
```
vaikas, line 31: this is neat, for the Rabbit cluster operator, I needed to spin through all the conditions since they didn't have a top level condition. It might be useful to have another variant of this (or not) that says, if all the conditions are ready, or be able to pass an array of conditions that must be ready for the resource to be ready. obvs, followup if ever.

n3wscott: I was thinking we could encode the logic like that for non-standard resources in here somewhere. I don't think it is a bad idea! it would be nice to just ask if any resource is ready and get an answer

PR #61, opened 2020-11-24
Add utility to return the address of an Addressable.

n3wscott:
# Changes

- `k8s.Address()` returns the url of an addressable.
- `k8s.IsAddress()` uses `k8s.Address()`
- New entry for eventshub: `StartSenderURL`, to pass the sink directly as a url.

**Release Note**

```release-note
Adding `k8s.Address()`, it returns the url of an addressable.
Adding `eventshub.StartSenderURL()`, to pass the sink directly as a url.
```

### DIFF
--- a/pkg/eventshub/options.go
+++ b/pkg/eventshub/options.go
@@ -45,6 +45,15 @@ func StartSender(sinkSvc string) EventsHubOption {
 	})
 }
 
+// StartSenderURL starts the sender in the eventshub sinking to a URL.
+// This can be used together with InputEvent, AddTracing, EnableIncrementalId, InputEncoding and InputHeader options
+func StartSenderURL(sink string) EventsHubOption {
+	return compose(envAdditive("EVENT_GENERATORS", "sender"), func(ctx context.Context, envs map[string]string) error {
+		envs["SINK"] = sink
+		return nil
+	})
+}
+
 // --- Receiver options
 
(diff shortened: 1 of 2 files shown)


pkg/eventshub/options.go
```diff
@@ -45,6 +45,15 @@ func StartSender(sinkSvc string) EventsHubOption {
...
 
+// StartSenderURL starts the sender in the eventshub sinking to a URL.
+// This can be used together with InputEvent, AddTracing, EnableIncrementalId, InputEncoding and InputHeader options
+func StartSenderURL(sink string) EventsHubOption {
+	return compose(envAdditive("EVENT_GENERATORS", "sender"), func(ctx context.Context, envs map[string]string) error {
+		envs["SINK"] = sink
+		return nil
+	})
+}
+
```
slinkydeveloper, line 56: A start sender version to an addressable?

slinkydeveloper (approved):
/lgtm
Just a comment for a followup

pkg/eventshub/options.go
n3wscott: That would be ideal. I think we need to change this to K_SINK and allow the sender pod to crash when K_SINK is not present and then we can use SinkBinding.

slinkydeveloper: If `sender` is set and there is no `SINK`, eventshub should already crash

Issue #31, opened 2020-11-11
complementary to flags, can we use some yaml?


n3wscott:
I wonder, more than flags (or complementary to flags), can we use some yaml config files? We have so many stuff we're testing and flags are becoming really messy and hard to read https://github.com/knative/eventing/blob/master/test/e2e-conformance-tests.sh#L36.

Even env variables looks better, we could have a script to source to load them before starting `go test`

_Originally posted by @slinkydeveloper in https://github.com/knative-sandbox/reconciler-test/pull/30#discussion_r520364692_

Issue #66, opened 2020-11-25
Framework needs to help test authors understand how to pass state between steps


n3wscott:
One issue has the testing framework makes it had to pass state between components. How do we solve this? Is it a real problem?

slinkydeveloper:
It's a real problem, we need to find a solution for this

n3wscott:
```
f.Setup("install source", func(ctx context.Context, t *testing.T) {
	// to a bunch of steps that create a thing and it eventually gets a status.address `fooAddress`
  // ...but this might not happen quickly.
}.
Setup("install source", func(ctx context.Context, t *testing.T) {
	// Use that resulting `fooAdddress`
	// This will never work at the moment.
})
```

OK brain wave....

It was my intention that the requirement timing would let you block until it passes while setup phases were happening so the testing phase works.

BUT! what if we change how Requirement works, and it gets paired with Setup as a prereq for running Setups. The setup phase is optional, example:

```
f.Setup("install foo", func(ctx context.Context, t *testing.T) {
	// to a bunch of steps that create a thing and it eventually gets a status.address `fooAddress`
  // ...but this might not happen quickly.
}.
Require("install thing targeting foo", func(ctx context.Context, t *testing.T) {
	// Wait for `fooAdddress` to be on the target resource.
}).
Setup("install thing targeting foo", func(ctx context.Context, t *testing.T) {
	// Use that resulting `fooAdddress`
})
Require("wait for all things to be ready", func(ctx context.Context, t *testing.T) {
	// this is an example of something that has no setup pair.
})
```

How to know a Require is paired with a Setup? Match the name of the step.

ok this does not solve the problem at hand, so I think we also introduce a config store that can be blocking:

```
f.Setup("install foo", func(ctx context.Context, t *testing.T) {
	// to a bunch of steps that create a thing and it eventually gets a status.address `fooAddress`
  // ...but this might not happen quickly.
}.
Require("install thing targeting foo", func(ctx context.Context, t *testing.T) {
	// Wait for `fooAdddress` to be on the target resource.
  fooAddress := "some valid value pulled from the cluster."
  f.SetConfig(ctx, "fooAddress", fooAddress)
}).
Setup("install thing targeting foo", func(ctx context.Context, t *testing.T) {
	// Use that resulting `fooAdddress`
  f.GetConfig(ctx, "fooAddress", &fooAddress) // maybe a return error, timeout, maybe it blocks? idk yet.
})
[...]

PR #82, opened 2021-02-09
Feature config

slinkydeveloper:
Fix https://github.com/knative-sandbox/reconciler-test/issues/31

# Changes

- :gift: Now you can configure all the features via config files

Fixes #

**Release Note**

```release-note
:gift: Now you can configure all the features via config files
```


pkg/config/config.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path"
+
+	"sigs.k8s.io/yaml"
+)
+
+var globalConfig map[string][]byte
```
dprotaso, line 15: no globals please

dprotaso: I wonder if this should part of the `Environment`

slinkydeveloper: How do you propose to implement it instead? Note that by design the config reading is one per process and I expect the user to invoke it in `TestMain`

slinkydeveloper: > I wonder if this should part of the Environment

Sounds like a good idea

slinkydeveloper: Nono i'm sorry, just checked, it doesn't work using env, because a feature can be built without the env

slinkydeveloper: So I believe we should leave it as a global as is now

pkg/config/config.go
```diff
@@ -0,0 +1,65 @@
...
+
+// ReadConfigNamed is like ReadConfig, but you can specify a file name
+func ReadConfigNamed(name string) {
```
n3wscott, line 18: instead of the global, you can return the parsed config.

pkg/feature/config.go
```diff
@@ -0,0 +1,16 @@
+package feature
+
+import (
+	"fmt"
+
+	"knative.dev/reconciler-test/pkg/config"
+)
+
+func (f *Feature) Config(out interface{}) interface{} {
+	err := config.UnmarshalConfig(f.Name, out)
```
n3wscott, line 10: just name might not be enough to know which config to load. For example, "BrokerTest" is likely going to be something that is used a few times?

n3wscott:
I do wonder how this would work with features that vendor and extend. Which leads me to think the config should always be passed to the feature constructor and not as the example shows where the feature constructor pulls out of global config.

pkg/feature/config.go
slinkydeveloper: My expectation is that test names are unique. I'm thinking about a "nested" configuration, but as a next step

slinkydeveloper:
> Which leads me to think the config should always be passed to the feature constructor and not as the example shows where the feature constructor pulls out of global config.

My expectation is that the user can configure per feature the various parameters, eg in this test https://github.com/knative/eventing/blob/master/test/rekt/features/broker_feature.go#L37 one more than passing the parameters, just gets the conf, corresponding to the feature name, and parse it to read the broker class

pkg/config/config.go
slinkydeveloper: That's not the way i thought to configuration. I would love to have a global conf for the whole test suite and then single "feature factories" read that config as they want.

pkg/config/config.go
dprotaso: `because a feature can be built without the env`

But a feature is executed against an environment - and technically are we not 'configuring' the feature in order for the test to pass?

I feel like this is related to: https://github.com/knative-sandbox/reconciler-test/issues/66

slinkydeveloper: > technically are we not 'configuring' the feature in order for the test to pass?

But the configuration happens before the test runs

slinkydeveloper:
Ok I feel like I misunderstood how we expect users to configure features:

* Do we expect _Feature constructors_ to have parameters, so vendors assemble their test package and execute the tests?
* Do we expect _Feature constructors_ to be no-arg, and then vendors can just grab the test executable and replace the `config.yaml` (this PR goes in this direction)

What are your opinions @dprotaso @n3wscott?

n3wscott:
I was expecting feature constructors to take in args when the test they do can be or wants to be parameterized. I think the author of the test needs to consider how the test is extended and vendored.

slinkydeveloper:
> I was expecting feature constructors to take in args when the test they do can be or wants to be parameterized. I think the author of the test needs to consider how the test is extended and vendored.

If that's the case, then this PR doesn't make any sense. I'll close it

pkg/config/config.go
n3wscott: the plan was to have that but do it in the testing domain, like eventing or serving and not the testing framework

PR #97, opened 2021-02-17
Refactoring feature steps execution

slinkydeveloper:
This PR revisits the way tests are executed and introduces some constraint on what can be executed inside the `StepFn`.

## The `feature.T` interface

The most important (and breaking) change is the `feature.T` interface. A `StepFn` now can fail, err, skip and log but it cannot create a subtest nor do other fancy stuff. This constrain a bit the step capabilities, in order to "force" users to create different feature steps. This also simplifies the assumptions we can do on whether happens in a step fn and allows us to record the stepfn results independently from what go testing is doing.

## The new step fn execution

Now the step fn (you can check it out in `magic.go`) doesn't run anymore in `t.Run`, but we manually run the stepFn in a specific goroutine that makes sure:

* context is properly handled per step-fn
* steps are executed in order
* one step fn cannot influence the execution of other step fn

I believe this approach gives us greater flexibility (e.g. we might have a policy that allows a test to not fail, after a feature step fails) and removes some race conditions (like https://github.com/knative/eventing/issues/4637 and the https://github.com/knative-sandbox/reconciler-test/pull/52).

To make sure our infra understands the results, I've adapted the output to the go test output. We could also work on a different "test results output format" in future that better adapts to our needs. E.g. we could output some "table" of features and, for each feature, what's the "reached" compliance level. Example: I have the retry feature, and say there is a SHOULD and a MUST. I want in output some "table" that tells me "your component retry feature compliance is at level MUST", even better it would be nice to save it in some file where I/the CI can grab it and process it.

### DIFF
--- a/pkg/environment/magic.go
+++ b/pkg/environment/magic.go
@@ -19,7 +19,9 @@ package environment
 import (
 	"context"
 	"fmt"
+	"strings"
 	"testing"
+	"time"
 
 	"go.uber.org/atomic"
 	corev1 "k8s.io/api/core/v1"
@@ -136,27 +138,32 @@ func (mr *MagicEnvironment) Test(ctx context.Context, t *testing.T, f *feature.F
 
 	steps := feature.ReorderSteps(f.Steps)
 
+	featureTestName := strings.ReplaceAll(f.Name, " ", "_")
 	for _, s := range steps {
-		fmt.Printf("=== %s   %s/%s/%s\n", run, t.Name(), f.Name, s.TestName())
+		stepName := strings.ReplaceAll(s.TestName(), " ", "_")
+		fmt.Printf("=== %s   %s/[%s]%s\n", run, t.Name(), featureTestName, stepName)
 
-		skipped, failed := mr.safeExecuteStep(ctx, t, s)
+		skipped, failed, duration := mr.safeExecuteStep(ctx, t, s)
 
 		if skipped {
-			fmt.Printf("--- %s   %s/%s/%s\n", skip, t.Name(), f.Name, s.TestName())
+			fmt.Printf("    --- %s: %s/[%s]%s (%.2fs) \n", skip, t.Name(), featureTestName, stepName, duration.Seconds())
 		} else if failed {
-			fmt.Printf("--- %s   %s/%s/%s\n", fail, t.Name(), f.Name, s.TestName())
+			fmt.Printf("    --- %s: %s/[%s]%s (%.2fs) \n", fail, t.Name(), featureTestName, stepName, duration.Seconds())
 			t.FailNow() // Here we can have different policies, depending on feature level etc
+		} else {
+			fmt.Printf("    --- %s: %s/[%s]%s (%.2fs) \n\n", pass, t.Name(), featureTestName, stepName, duration.Seconds())
 		}
 	}
 }
 
 const (
 	run  = "RUN"
+	pass = "PASS"
 	fail = "FAIL"
 	skip = "SKIP"
 )
 
-func (mr *MagicEnvironment) safeExecuteStep(ctx context.Context, testingT *testing.T, step feature.Step) (skipped bool, failed bool) {
+func (mr *MagicEnvironment) safeExecuteStep(ctx context.Context, testingT *testing.T, step feature.Step) (skipped bool, failed bool, duration time.Duration) {
 	testingT.Helper()
 
 	if mr.s&step.S == 0 || mr.l&step.L == 0 {
@@ -185,6 +192,8 @@ func (mr *MagicEnvironment) safeExecuteStep(ctx context.Context, testingT *testi
 		ctx, cancelFn = context.WithCancel(ctx)
 	}
 	var panicValue interface{}
+
+	start := time.Now()
 	go func() {
 		testingT.Helper()
 		defer func() {
@@ -199,6 +208,7 @@ func (mr *MagicEnvironment) safeExecuteStep(ctx context.Context, testingT *testi
 	}()
 
 	<-ctx.Done()
+	duration = time.Now().Sub(start)
 	if ctx.Err() == context.DeadlineExceeded {
 		testingT.Logf("Timeout exceeded")
 		t.failed.Store(true)
--- a/pkg/environment/magic.go
+++ b/pkg/environment/magic.go
@@ -133,9 +133,14 @@ func (mr *MagicEnvironment) Prerequisite(ctx context.Context, t *testing.T, f *f
 func (mr *MagicEnvironment) Test(ctx context.Context, t *testing.T, f *feature.Feature) {
 	t.Helper() // Helper marks the calling function as a test helper function.
 
-	steps := feature.CollapseSteps(f.Steps)
+	steps := feature.ReorderSteps(f.Steps)
+
+	for _, s := range steps {
+		t.Run()
+	}
 
 	for _, timing := range feature.Timings() {
+
 		// do it the slow way first.
 		wg := &sync.WaitGroup{}
 		wg.Add(1)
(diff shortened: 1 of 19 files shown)


slinkydeveloper:
cc @n3wscott @dprotaso @vagababov

pkg/environment/magic.go
```diff
@@ -133,46 +136,92 @@ func (mr *MagicEnvironment) Prerequisite(ctx context.Context, t *testing.T, f *f
...
-		})
+			cancelFn()
+		}()
+		step.Fn(ctx, &t)
+	}()
+
+	<-ctx.Done()
+	duration = time.Now().Sub(start)
+	if ctx.Err() == context.DeadlineExceeded {
+		testingT.Logf("Timeout exceeded")
```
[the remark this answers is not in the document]
dprotaso: Also given [`-timeout` panics](https://golang.org/cmd/go/#hdr-Testing_flags)
>  If a test binary runs longer than duration d, panic.

I wonder if we can _even_ this clean up will occur

pkg/feature/t.go
```diff
@@ -0,0 +1,18 @@
+package feature
+
+type T interface {
```
dprotaso, line 3: There's probably more that are useful

ie. Cleanup,Deadline,Helper etc.

dprotaso (commented):
> skip and log but it cannot create a subtest nor do other fancy stuff. 

why do you want to restrict it?

pkg/environment/magic.go
```diff
@@ -133,46 +136,92 @@ func (mr *MagicEnvironment) Prerequisite(ctx context.Context, t *testing.T, f *f
...
+			if panicValue = recover(); panicValue != nil {
+				t.failed.Store(true)
 			}
-		})
+			cancelFn()
+		}()
+		step.Fn(ctx, &t)
+	}()
+
+	<-ctx.Done()
```
dprotaso, line 210: Since this waits for the test to finish why use a go routine at all?

pkg/environment/magic.go
```diff
@@ -133,46 +136,92 @@ func (mr *MagicEnvironment) Prerequisite(ctx context.Context, t *testing.T, f *f
...
-					wg.Add(1)
-					defer wg.Done()
+		if skipped {
+			fmt.Printf("    --- %s: %s/[%s]%s (%.2fs) \n\n", skip, t.Name(), featureTestName, stepName, duration.Seconds())
+		} else if failed {
+			fmt.Printf("    --- %s: %s/[%s]%s (%.2fs) \n\n", fail, t.Name(), featureTestName, stepName, duration.Seconds())
+			t.FailNow() // Here we can have different policies, depending on feature level etc
+		} else {
+			fmt.Printf("    --- %s: %s/[%s]%s (%.2fs) \n\n", pass, t.Name(), featureTestName, stepName, duration.Seconds())
+		}
```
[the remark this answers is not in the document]
dprotaso: re: mutliple features being executed

https://github.com/knative-sandbox/reconciler-test/pull/102

pkg/environment/t.go
```diff
@@ -0,0 +1,74 @@
+package environment
+
+import (
+	"runtime"
+	"testing"
+
+	"go.uber.org/atomic"
+)
+
+type t struct {
```
n3wscott, line 10: Why is this needed?

pkg/feature/t.go
n3wscott: I really don't want to define this interface.

pkg/feature/feature.go
```diff
@@ -29,7 +28,7 @@ type Feature struct {
 }
 
 // StepFn is the function signature for steps.
-type StepFn func(ctx context.Context, t *testing.T)
+type StepFn func(ctx context.Context, t T)
```
n3wscott, line 31: hard no thank you on this one. The whole point of using `*testing.T` is I can define a set of steps and not depend on this library at all. I don't want to force an interface here, I really just want it to be a standard testing.T object.

pkg/environment/magic.go
```diff
@@ -133,46 +136,92 @@ func (mr *MagicEnvironment) Prerequisite(ctx context.Context, t *testing.T, f *f
...
-					// Perform step.
-					s.Fn(ctx, t)
+	t := t{
+		t:       testingT,
+		failed:  atomic.NewBool(false),
+		skipped: atomic.NewBool(false),
+	}
 
-				})
+	var cancelFn context.CancelFunc
```
n3wscott, line 187: This all looks very complicated and I don't see why it would be better than letting the testing lib do this work.

n3wscott (changes requested):
There are several improvements we should cherrypick out of this PR but overall I really really really do not want to change the signature for `StepFn` to force providers of StepFn impls to need to vendor this lib. 

I think the timing things are very complicated and I have not seen a need to do more than what standard golang testing is doing today.

Controlling the output seems like a direct user feedback thing that is nice but not necessary. 

I would rather think about how to collect the results and then provide them as a file and/or printed at the end of the test run to give tooling a structured way to look at results over trying to make the output easier to read for a human.

Overall, our goal here is not to be fancy, we need to focus on the UX of writing tests and having them be as vendorable, composable, and repeatable as possible while not needing to teach any new ways to control tooling or significantly change how the code is written. 

Better or worse, go choose to use the struct `*testing.T` and I really do not want to be in the business of wrapping, shimming or playing catchup as new features come from that area.

slinkydeveloper:
Ok so I admit there are some several points in this PR that I got in my mind while working on it, so let me try to go through it step by step:

> I really don't want to define this interface.

About the `feature.T` interface, the main reason behind it is to avoid users to do `t.RunTest`. First and most important, `RunTest` inside a `StepFn` doesn't make sense, or you have different `Assert` steps or you have different `Feature`s.
Also, if you execute `t.RunTest` inside a StepFn, it can break everything regarding . go testing doesn't care about setup, requirement, assert and teardown, even less it cares about our context, our execution order etc. https://stackoverflow.com/questions/31201858/how-to-run-golang-tests-sequentially . Since today we have the "steps collapse" I've implemented (which might be wrong and might not really solve cleanly the original problem), If I run inside a `Setup` a `t.RunTest`, the execution of the subtest might actually end after the callee `StepFn` completed the execution.

With this interface (which is drafted, i know, and it definitely requires Cleanup etc) we still have a `testing.T` compliant type, so it works with various `testify` and friends and, at the same time, forces our restrictions and gives us greater flexibility in future to change the underlying test execution code. We also have the control on test results and, thanks to that, we can have a custom report of the test results, like explained in the PR description. Having a detailed "compliance report" is one of the most important features we initially identified for this framework, and IMO parsing the test output will never give us the level of detail we want.
In other words, exposing `testing.T` in `StepFn` to me is like exposing the implementation detail that this library is relying on `go test` and this doesn't give us flexibility to implement all the things we want.

The test execution with all the various logging statements etc is a separate point from the `feature.T` interface (they can co-exist, or just `feature.T` can exist). I wrongly put both in the same PR. The idea of rewriting the test execution code was to force exactly our "guarantees" of the execution order, define specific policies when some phase fails. e.g. if a requirement or setup phase fails, the execution of the feature should not continue, but if a requirement step fail, maybe the test should not be flagged as failed but more as a skipped with the failure reason from the requirement phase (which brings me to the previous point of having a `feature.T` interface). Since go test doesn't give us the flexibility of failing/skipping/controlling the execution of tests directly based on the result of others, I tried to rewrite the bits from `testing` that we need to execute the testing code here

> to force providers of StepFn impls to need to vendor this lib.

I don't understand why this argument is a problem, since the user of this framework and of StepFn is going to depend on this framework anyway.

pkg/feature/t.go
n3wscott: Ok, I talked offline with Slinky and we are going to keep the interface.

slinkydeveloper:
I'm closing this one since it's a mix of things and splitted the discussion:

* https://github.com/knative-sandbox/reconciler-test/issues/106
* https://github.com/knative-sandbox/reconciler-test/pull/105

pkg/feature/feature.go
n3wscott: we are going to try it.

Issue #65, opened 2020-11-25
Framework needs a way to group Features into Sets


n3wscott:
Two things will need to be added, something like:

1. feature.Set is introduced, it is a list of `feature.Feature`s that run as a unit. 
2. `env.TestSet` is added as a test entry point to run these sets of features.

The use case here is something like a conformance suite that would like to bundle the artifact of N features as a unit and have that be vendored downstream.

PR #102, opened 2021-02-23
Adding FeatureSet and TestSet

n3wscott:
# Changes

fixes https://github.com/knative-sandbox/reconciler-test/issues/65

- :gift: Adding FeatureSet, a wrapper around a set of features.
- :gift: Adding TestSet, a test entry point to run a FeatureSet.

/kind enhancement

We need a way to vendor a group of features rather than individuals, the aim here is to provide a set of features that might be required for something like a conformance test, and it does not make much sense to vendor them individually as a burden to the downstream user of these.

**Release Note**

```release-note
Adding feature.FeatureSet and Environment.TestSet.
```

### DIFF
--- a/pkg/environment/magic.go
+++ b/pkg/environment/magic.go
@@ -184,11 +184,12 @@ func (mr *MagicEnvironment) TestSet(ctx context.Context, t *testing.T, fs *featu
 
 	// do it the slow way first.
 	wg := &sync.WaitGroup{}
-	wg.Add(1)
 
 	for _, f := range fs.Features {
+		wg.Add(1)
 		t.Run(fs.Name, func(t *testing.T) {
 			mr.Test(ctx, t, &f)
+			wg.Done()
 		})
 	}
 
--- a/pkg/environment/magic.go
+++ b/pkg/environment/magic.go
@@ -64,7 +64,9 @@ func (mr *MagicEnvironment) References() []corev1.ObjectReference {
 }
 
 func (mr *MagicEnvironment) Finish() {
-	mr.DeleteNamespaceIfNeeded()
+	if err := mr.DeleteNamespaceIfNeeded(); err != nil {
+		panic(err)
+	}
 }
 
 func (mr *MagicGlobalEnvironment) Environment(opts ...EnvOpts) (context.Context, Environment) {
@@ -136,7 +138,6 @@ func (mr *MagicEnvironment) Test(ctx context.Context, t *testing.T, f *feature.F
 	steps := feature.CollapseSteps(f.Steps)
 
 	for _, timing := range feature.Timings() {
-		// do it the slow way first.
 		wg := &sync.WaitGroup{}
 		wg.Add(1)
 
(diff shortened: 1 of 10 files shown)


vaikas:
Looks like a very nice way to group features!

pkg/environment/magic.go
```diff
@@ -178,6 +178,24 @@ func (mr *MagicEnvironment) Test(ctx context.Context, t *testing.T, f *feature.F
 	}
 }
 
+// TestSet implements Environment.TestSet
+func (mr *MagicEnvironment) TestSet(ctx context.Context, t *testing.T, fs *feature.FeatureSet) {
+	t.Helper() // Helper marks the calling function as a test helper function.
+
+	// do it the slow way first.
```
vaikas, line 185: is there a TODO here?

pkg/environment/magic.go
```diff
@@ -178,6 +178,24 @@ func (mr *MagicEnvironment) Test(ctx context.Context, t *testing.T, f *feature.F
...
+	wg := &sync.WaitGroup{}
+
+	for _, f := range fs.Features {
+		wg.Add(1)
+		t.Run(fs.Name, func(t *testing.T) {
+			mr.Test(ctx, t, &f)
```
vaikas, line 191: does this need t.Parallel() here, or is it done elsewhere? 
https://golang.org/pkg/testing?

dprotaso:
You should probably review https://github.com/knative-sandbox/reconciler-test/pull/97 first as it has assumptions that are incompatible with this PR

pkg/environment/magic.go
n3wscott: 👍 Yes, I need to follow up on this.

pkg/environment/magic.go
n3wscott: that was just notes left behind. `rm -rf`

pkg/environment/magic.go
n3wscott: It turns out, we already do this: 

```
func TestFoo(t *testing.T) {
	// Signal to the go test framework that this test can be run in parallel
	// with other tests.
	t.Parallel()
```

The parent test controls if all the following will be in parallel or not. so we are good.

vaikas (approved):

vaikas:
/hold
hold since Dave had comments

pkg/environment/magic.go
```diff
@@ -178,6 +179,23 @@ func (mr *MagicEnvironment) Test(ctx context.Context, t *testing.T, f *feature.F
...
+	for _, f := range fs.Features {
+		wg.Add(1)
+		t.Run(fs.Name, func(t *testing.T) {
+			// FeatureSets should be run in parellel.
+			mr.Test(ctx, t, &f)
+			wg.Done()
+		})
+	}
+
+	wg.Wait()
```
dprotaso, line 196: This will deadlock

see: https://play.golang.org/p/rXUUCO1bUha

Essentially when you call `t.Parallel` child tests wait for the top level test to finish before continuing. There's no way to wait for the children to finish

dprotaso: nm - i had a bug with my use of defer

dprotaso: Ok it can still happen

https://play.golang.org/p/OEG1zc6cN-N

n3wscott: ok we can fix this here and everywhere else in a followup

PR #103, opened 2021-02-24
enable saving config between steps in a single feature

n3wscott:
# Changes

Looking for feedback on a method to pass data between steps in a feature.

Note that the interface introduced here is aligned with the implementation in pkg: https://github.com/knative/pkg/tree/master/kvstore

/kind enhancement

Relates to https://github.com/knative-sandbox/reconciler-test/issues/66


pkg/feature/feature.go
```diff
@@ -52,6 +55,22 @@ func (s *Step) TestName() string {
 	}
 }
 
+// Save a value for use in a later feature step, or error on issue.
+func (f *Feature) Save(ctx context.Context, t *testing.T, key string, value interface{}) {
```
vaikas, line 59: Just curious what the reasoning is for Save/Load vs. Set/Get? I found it a bit odd that we translate Save/Load into Get/Set underneath.

vaikas: I guess my thinking is that to me Save/Load semantically means save everything or load everything.

pkg/state/interfaces.go
```diff
@@ -0,0 +1,27 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package state
+
+import "context"
+
+// This is a subset of knative.dev/pkg/kvstore.Interface
+type Store interface {
```
vaikas, line 22: Instead of defining a separate interface, would it be bad if we did not define this new interface, but instead in the InMemory implementation, we return NonImplemented error instead?

pkg/feature/feature.go
n3wscott: I was trying to make a helper method for getting and setting the keys tied up with failing the test, so maybe you have a better name for the method? Maybe I also call it Get and Set, note it is on the top level object, not the Store object.

pkg/state/interfaces.go
n3wscott: I was attempting to allow to use one of those impls, but not vendor it in directly because the Load and Init and Save did not make sense for Features

slinkydeveloper (commented):
I like the approach overrall, but i don't like that i need to use `feature` to `Load` and `Store`. This approach with feature is not really modular, because you can't provide a stepfn that wants to push things inside the store without without knowing in advance the feature, breaking the encapsulation between feature and stepfn (for example, think about one of the recorder features pushing useful data inside the kv store when installing for later usage)

What I had in mind was something using the context, where the kv store lives in the context and this context is passed through the various phases of the execution:

```
	f.Setup("install sender", func(ctx context.Context, t *testing.T) {
		t.Helper()
		var to string
		var event cloudevents.Event
		storepkg.Load(ctx, t, "to", &to)
		storepkg.Load(ctx, t, "event", &event)
		from := feature.MakeRandomK8sName("sender")
		eventshub.Install(from, eventshub.StartSender(to), eventshub.InputEvent(event))
	})
``` 

(`storepkg` is the module where kvstore impl lives)

Also you should provide a `Get` method that returns `interface{}` too, so user can choose between the two "styles"

n3wscott:
having the store be context based will not work because of how passing and threading work, it would have to be injected into the context higher up and then used as a pointer to another kvstore, which we can do as a followup that is feature based. So I say we start with this when I fix it up with the comments and we can look at fancy hooks to access the kvstore later? or maybe ... meh YOLO, brb.

n3wscott:
ok please take another look, I took the request from @vaikas and @slinkydeveloper and made the feature integration look like this:

setup,
```
		state.SetOrFail(ctx, t, "event", FullEvent())
...
		state.SetOrFail(ctx, t, "to", to)
```

using, 
```
		to := state.GetStringOrFail(ctx, t, "to")
		var event cloudevents.Event
		state.GetOrFail(ctx, t, "event", &event)
```

but nothing is stopping you from loading the KVStore directly:

```
s := state.FromContext(ctx)
s.Get/Set ...
```

vaikas:
/hold
Just holding in case @slinkydeveloper has more thoughts.
But otherwise, LGTM. Thanks!

vaikas (dismissed):

pkg/state/context.go
```diff
@@ -0,0 +1,62 @@
...
+	"testing"
+)
+
+type envKey struct{}
+
+func ContextWith(ctx context.Context, s Store) context.Context {
+	return context.WithValue(ctx, envKey{}, s)
+}
+
+func FromContext(ctx context.Context) Store {
```
slinkydeveloper, line 30: Please add a comment here that makes explicit that the KVStore you get from this method is "feature" scoped

pkg/state/kvstore.go
```diff
@@ -0,0 +1,67 @@
...
+	if cs.store == nil {
+		cs.store = make(map[string]string)
+	}
+
+	v, ok := cs.store[key]
+
+	if !ok {
+		return fmt.Errorf("key %s does not exist", key)
+	}
+	err := json.Unmarshal([]byte(v), value)
```
slinkydeveloper, line 45: Ah wait i didn't saw it, why storing json? Can't we store just void pointers to things and then cast them when retrieving them? The state to store doesn't necessarily define conversion back and forth to json, and for some reason the user might even not be able to define such conversion (e.g. when you're using a struct from an external library as state)

n3wscott: I avoided this way to prevent mutation of the thing that is stored in the kvstore. Set, Get, mutate get'ed object, Get updated object?? is unexpected to me. cheap and cheerful way is json.

pkg/state/context.go
n3wscott: will do.

n3wscott: I added a comment for context in MagicEnvironment.Test because that is where this assumption is happening, this method just reaches into context and pulls out the last Store. You can use that for anything you want.

n3wscott:
Ready.

/unhold

pkg/state/context.go
```diff
@@ -23,10 +23,15 @@ import (
...
 
+// ContextWith decorates the given context with the provided Store, and returns
+// the resulting context.
 func ContextWith(ctx context.Context, s Store) context.Context {
 	return context.WithValue(ctx, envKey{}, s)
 }
 
+// FromContext returns the Store from Context, if not found FromContext will
+// panic.
+// TODO: revisit if we really want to panic here... likely not.
```
slinkydeveloper, line 34: IMO panic is absolutely fine here, the `Store` must always exists in the context passed to `StepFn`. If it's not there, it's a reconciler-test bug

pkg/state/kvstore.go
slinkydeveloper: I see your point, but maybe what I want to store inside the kvstore is exactly a mutable object, like a list.

In any case, my problem with json is that you might end up storing objects which are not very json compatible but where marshalling doesn't fail, and you will lose some information, and together with it debugging time.

pkg/state/kvstore.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to Unmarshal %q: %v", v, err)
+	}
+	return nil
+}
+
+// Set marshals and sets the value given under specified key.
+func (cs *KVStore) Set(_ context.Context, key string, value interface{}) error {
```
slinkydeveloper, line 53: Do we expect users to check the `error` everytime and fail the `StepFn` if the `Set`/`Get` returns an error? Isn't that too much and makes the usage unnecessary complicated? Can't we just panic in case of an error? These `Get`/`Set` errors are not recoverable anyway.

n3wscott: I am matching a subset of this interface: https://github.com/knative/pkg/blob/master/kvstore/kvstore.go#L23

pkg/state/interfaces.go
```diff
@@ -0,0 +1,27 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package state
+
+import "context"
+
+// This is a subset of knative.dev/pkg/kvstore.Interface
```
n3wscott, line 21: note: @slinkydeveloper ^^

pkg/state/kvstore.go
n3wscott: the impl can evolve over time, this is more about the interface and the current problemset is really "how do I pass that url to the next phase" 

if you start working out ways of passing callbacks I will burn the building down.

pkg/state/interfaces.go
slinkydeveloper: Any particular reason why we're trying to match that particular interface about config maps?

pkg/state/kvstore.go
slinkydeveloper: Since it's an impl detail, i guess you can always have the `GetOrFail` and `SetOrFail`

pkg/state/kvstore.go
slinkydeveloper: `net.Url` is a too simplistic example (and even with it, sometimes encoding and decoding it back and forth to json has weird behaviours). Any more complex state where json encoding/decoding is not defined won't work with this implementation.  

> if you start working out ways of passing callbacks I will burn the building down.

That's state of the test too :smile:

n3wscott: it is a kv store, if you can't save it to a config map, it will not work. That is the bar.

pkg/state/kvstore.go
n3wscott: We can't do that because KVStore is not used, it is Store, and Store is the pkg kvstore interface.

I have added methods like this to the context.

pkg/state/interfaces.go
n3wscott: So that we can swap out this impl for the same that is used in pkg.

n3wscott: We are trying to make synergy here, not reboil the world. I made the interface a subset because some of those things did not make sense in the context of features, but I want to use that impl in pkg at some point later.

pkg/state/kvstore.go
slinkydeveloper: But why are we using the kv store model for something which is always in memory by design? It's an in memory map, it's not something that has to be read/write to a config map.

n3wscott: it might let us do some really interesting things later, like retries or test startup in the middle, or capture test state.

vaikas (approved):

Issue #106, opened 2021-03-02
Test execution model


slinkydeveloper:
Today we have this code https://github.com/knative-sandbox/reconciler-test/blob/master/pkg/environment/magic.go where we take the various user `StepFn`s and execute them.

This model has several problems:

* We don't really implement `Requirement` as we would love to: if a `Requirement` fails, we don't want the test to fail, but we want to skip the asserts (don't execute them at all) and just run the teardown. More generally, we cannot really control what to do after a StepFn fails or succeds, which is preferable given our APIs and the meanings of our "phases".
* Test branching `RunTest` is allowed but it doesn't really makes sense to do so, and in setup, teardown and requirement phases it doesn't work properly.
* Context is not properly handled per `StepFn` (I attempted to fix it here https://github.com/knative-sandbox/reconciler-test/pull/96)
* Test phases execution is enforced through running all the phases in the same subtest (as implemented by me in https://github.com/knative-sandbox/reconciler-test/pull/52), but this solution breaks subtests, panic catching, cleanup, skipping and deadline features of `*testing.T`
* How `MustNot` and `ShouldNot` be implemented? The user should assert the original or negated assertion?

I attempted to fix it by reimplementing part of the job of `go test` in this PR https://github.com/knative-sandbox/reconciler-test/pull/97 but there are several issues in it, most notably the fact that I'm reimplementing `go test` :smile: 

cc @dprotaso @n3wscott do you have any viable solutions to fix the above issues?

dprotaso:
The other notable thing I realized today is `go test ./...` (when testing multiple packages) actually runs the test binaries in parallel.

so `go test pkgA/e2e pkgA/e2e/scale` will immediately spin up two processes. The # of processes is cpu count dependent.

There's actually a poorly documented flag that limits this parallelism `-p 1`. See: https://golang.org/cmd/go/#hdr-Compile_packages_and_dependencies

slinkydeveloper:
@dprotaso running parallel test binaries is not a big problem at the end of the day, it's better than running in parallel in the same process, which makes the whole execution model for us more complicated

dprotaso:
@slinkydeveloper I agree. I mentioned it since I'm poking at improving serving's e2e scripts.

The other notable thing I realize is that test timeouts panic so all the tear-downs/cleanup do not run. I think the only way to work around this is to wrap the test binary in a separate process in `-exec`. I'm looking into that now

slinkydeveloper:
[...]

PR #109, opened 2021-03-05
Adding milestone events to reconciler test framework.

n3wscott:
# Changes

I am thinking of ways to add observibility to the testing framework. At the same time, I was thinking about the  milestone events design doc for eventing and thought it would be a good fit here. The first implementation sends the events to an http endpoint, but you could very well add an implementation that defaults to writing to a file. The thinking here is two-fold:

- Enables realtime debugging abilities to the test run.
- Could be used to produce an audit log for processing and verification later (conformance, for example)

/kind enhancement

To see this in action, define an environment variable called `MILESTONE_EVENTS_TARGET` with a cloudevents consumer and run a test.

For example:

Install the [cloudevents conformance tool](https://github.com/cloudevents/conformance) and in another terminal, start it in `listen` mode:

```
$ cloudevents listen
``` 

Then run the tests:

```
MILESTONE_EVENTS_TARGET=http://localhost:8080 go test -v -count=1 -timeout=2m -tags=e2e ./test/...
```

This results in a stream of events that are major milestones in the testing suite, small example:

```yaml
ContextAttributes:
  specversion: "1.0"
  type: dev.knative.rekt.environment.v1
  time: "2021-03-05T00:18:51.995599Z"
  id: ffee45a9-10ab-4712-a74f-1d8879b4e3ec
  source: knative.dev/reconciler-test
  subject: /api/v1/namespaces/rekt-tqxnvuix
  datacontenttype: application/json
TransportExtensions:
  Accept-Encoding: gzip
  Content-Length: "75"
  User-Agent: Go-http-client/1.1
Data: '{"featureState":"Any","namespace":"rekt-tqxnvuix","requirementLevel":"All"}'
---
ContextAttributes:
  specversion: "1.0"
  type: dev.knative.rekt.namespace.created.v1
  time: "2021-03-05T00:18:51.90947Z"
  id: 525fda9f-ca89-441f-accd-8cf412b166a3
  source: knative.dev/reconciler-test
  subject: /api/v1/namespaces/rekt-tqxnvuix
  datacontenttype: application/json
TransportExtensions:
  Accept-Encoding: gzip
  Content-Length: "29"
  User-Agent: Go-http-client/1.1
Data: '{"namespace":"rekt-tqxnvuix"}'
---
ContextAttributes:
  specversion: "1.0"
  type: dev.knative.rekt.test.finished.v1
  time: "2021-03-05T00:19:12.24093Z"
  id: 22fa904c-cff1-4f32-b6fc-26a830785e51
  source: knative.dev/reconciler-test
  subject: /api/v1/namespaces/rekt-siqatkpw
  datacontenttype: application/json
  Extensions:
    Feature: EchoFeature
    Stepname: echo job is finished
    Testname: TestEchoSet/Echo_Feature_Wrapper_(3x)/Requirement/echo_job_is_finished
    Testparent: TestEchoSet
TransportExtensions:
  Accept-Encoding: gzip
  Content-Length: "188"
  User-Agent: Go-http-client/1.1
Data: '{"failed":false,"feature":"EchoFeature","passed":true,"skipped":false,"stepName":"echo
  job is finished","testName":"TestEchoSet/Echo_Feature_Wrapper_(3x)/Requirement/echo_job_is_finished"}'
---
ContextAttributes:
  specversion: "1.0"
  type: dev.knative.rekt.test.started.v1
  time: "2021-03-05T00:19:12.241598Z"
  id: bf703e81-666c-4ad1-9801-f493e08edbb1
  source: knative.dev/reconciler-test
  subject: /api/v1/namespaces/rekt-siqatkpw
  datacontenttype: application/json
  Extensions:
    Feature: EchoFeature
    Stepname: '[Alpha/MUST]pull logs off a pod the echo pod must echo our message'
    Testname: TestEchoSet/Echo_Feature_Wrapper_(3x)/Assert/[Alpha/MUST]pull_logs_off_a_pod_the_echo_pod_must_echo_our_message
    Testparent: TestEchoSet
TransportExtensions:
  Accept-Encoding: gzip
  Content-Length: "230"
  User-Agent: Go-http-client/1.1
Data: '{"feature":"EchoFeature","stepName":"[Alpha/MUST]pull logs off a pod the echo
  pod must echo our message","testName":"TestEchoSet/Echo_Feature_Wrapper_(3x)/Assert/[Alpha/MUST]pull_logs_off_a_pod_the_echo_pod_must_echo_our_message"}'
---
ContextAttributes:
  specversion: "1.0"
  type: dev.knative.rekt.test.started.v1
  time: "2021-03-05T00:19:12.242082Z"
  id: a791c100-7dd5-45f3-86ad-6d8c7ad3708b
  source: knative.dev/reconciler-test
  subject: /api/v1/namespaces/rekt-siqatkpw
  datacontenttype: application/json
  Extensions:
    Feature: EchoFeature
    Stepname: '[Alpha/MAY]pull logs off a pod An example of a MAY'
    Testname: TestEchoSet/Echo_Feature_Wrapper_(3x)/Assert/[Alpha/MAY]pull_logs_off_a_pod_An_example_of_a_MAY
    Testparent: TestEchoSet
TransportExtensions:
  Accept-Encoding: gzip
  Content-Length: "198"
  User-Agent: Go-http-client/1.1
Data: '{"feature":"EchoFeature","stepName":"[Alpha/MAY]pull logs off a pod An example
  of a MAY","testName":"TestEchoSet/Echo_Feature_Wrapper_(3x)/Assert/[Alpha/MAY]pull_logs_off_a_pod_An_example_of_a_MAY"}'
```

**Release Note**
```release-note
Adding milestone events to the rekt framework. Enabled by defining an environment variable MILESTONE_EVENTS_TARGET=<uri_to_cloudevents_comsume>.
```


slinkydeveloper:
The idea sounds interesting on the surface, but I don't get why:

* sending stuff to an external tool? Could these just be appended to an in memory map and then print it when the test fails?
* you're using cloudevents? (maybe that's a product of the first point)

Also I believe this relates to #106 because, depending on its outcome, this might be useful or not (what if we end up executing all the tests serially?)

n3wscott:
Cloudevents is the generic way to get those checkpoints out of the test runner without filling memory. 

If you run in parallel you can filter by test, namespace, or even testing instance run, and the next version of sockeye makes this easy.

pkg/environment/magic.go
```diff
@@ -22,25 +22,31 @@ import (
...
+		id:               uuid.New().String(),
 		RequirementLevel: *l,
 		FeatureState:     *s,
 	}
 }
 
 type MagicGlobalEnvironment struct {
 	c context.Context
+	// id is used to link runs together from a single global environment.
+	id string
```
slinkydeveloper, line 49: maybe rename it `runId` to clarify its scope?

pkg/environment/magic.go
```diff
@@ -55,8 +61,16 @@ type MagicEnvironment struct {
 	namespace        string
 	namespaceCreated bool
 	refs             []corev1.ObjectReference
+	// milestones sends milestone events, if configured.
+	milestones milestone.Sender
+	// mf is the milestone event factory.
+	mf *milestone.Factory
```
slinkydeveloper, line 67: rename `milestoneFactory`?

slinkydeveloper, line 67: I was wondering, in order to avoid creating events never used and just thrown away, what if we have a single interface, like `MilestoneEventEmitter`, which looks exactly like the factory methods, except it doesn't return an event but directly create and emit them? 

This interface won't be coupled to cloudevents, in fact the interface itself won't know nothing about cloudevents, so it could be more _related_ to the testing framework domain, allowing us to reuse it for other needs, like for logging etc. It also removes the need for the "interface user" (aka the code in magic.go) to have any knowledge at all about sending events.

pkg/environment/namespace.go
```diff
@@ -34,9 +34,9 @@ const (
 )
 
 // CreateNamespaceIfNeeded creates a new namespace if it does not exist.
-func (me *MagicEnvironment) CreateNamespaceIfNeeded() error {
-	c := kubeclient.Get(me.c)
-	nsSpec, err := c.CoreV1().Namespaces().Get(context.Background(), me.namespace, metav1.GetOptions{})
+func (mr *MagicEnvironment) CreateNamespaceIfNeeded() error {
```
slinkydeveloper, line 37: why the rename from mr to me?

pkg/environment/namespace.go
```diff
@@ -45,45 +45,47 @@ func (me *MagicEnvironment) CreateNamespaceIfNeeded() error {
...
+			return fmt.Errorf("the default ServiceAccount was not created for the Namespace: %s", mr.namespace)
 		}
 	}
 	return nil
 }
 
-func (me *MagicEnvironment) DeleteNamespaceIfNeeded() error {
-	if me.namespaceCreated {
-		c := kubeclient.Get(me.c)
+func (mr *MagicEnvironment) DeleteNamespaceIfNeeded() error {
```
slinkydeveloper, line 72: same

slinkydeveloper:
About feature naming: https://github.com/knative-sandbox/reconciler-test/pull/111

pkg/environment/magic.go
slinkydeveloper: Inside the implementation of such interface, say `CloudEventsMilestoneEventEmitter` you can then reuse the `milestone.Sender` interface you've used

n3wscott: mf = milestoneFactor, it is internal detail that I was trying to keep small.

n3wscott: Sure! That is interesting. Some of this is organic as I was making it...

pkg/environment/namespace.go
n3wscott: the rest of the receivers are `mr` in the other file and it makes me happy to type things like `mr.c`

pkg/environment/magic.go
n3wscott: I dropped the factory and it is now inside the sender.

n3wscott: done. check it!

pkg/milestone/sender.go
```diff
@@ -0,0 +1,181 @@
...
+	retryBackoff = 10 * time.Millisecond
+	maxTries     = 3
+)
+
+type envConfig struct {
+	MilestoneEventsTarget string `envconfig:"MILESTONE_EVENTS_TARGET"`
+}
+
+// Sender sends milestone events.
+type Sender interface {
```
slinkydeveloper, line 42: Emitter IMO is more appropriate and less _opinionated_ towards the actual impl

slinkydeveloper: Can we have a noop impl that just does nothing on each event?

pkg/milestone/sender.go
```diff
@@ -0,0 +1,181 @@
...
+// Sender to provide nil check safety.
+type NilSafeClient struct {
+	Client  cloudevents.Client
+	Factory *Factory
+}
+
+func (n *NilSafeClient) Environment(env map[string]string) {
+	if n == nil || n.Client == nil {
+		return
+	}
```
slinkydeveloper, line 107: Let's just have a noop impl more than this

pkg/milestone/sender.go
```diff
@@ -0,0 +1,181 @@
...
+	default:
+		return nil, fmt.Errorf("unsupported milestone event target uri: %q", target.String())
+	}
+}
+
+// NilSafeClient is a simple wrapper around a cloudevent client that implements
+// Sender to provide nil check safety.
+type NilSafeClient struct {
+	Client  cloudevents.Client
+	Factory *Factory
```
slinkydeveloper, line 101: No need to make these public

pkg/milestone/sender.go
```diff
@@ -0,0 +1,181 @@
...
+
+type envConfig struct {
+	MilestoneEventsTarget string `envconfig:"MILESTONE_EVENTS_TARGET"`
+}
+
+// Sender sends milestone events.
+type Sender interface {
+	// Send will send CloudEvents and return the result.
+	// Send must be implemented in a nil safe way.
+	Send(ctx context.Context, event cloudevents.Event)
```
slinkydeveloper, line 45: IMO the interface should not have this, cloudevents and the `Send` idiom are an implementation detail of your particular implementation of the event emitter. It doesn't make sense for this interface user to know about cloudevents, nor for some eventual implementations (like the logging one I proposed).
For example: if i would love to implement this `Emitter` for logging the "important" states of the test executions, what's the meaning of `Send`? And why is it important, in relation to the test framework?

pkg/milestone/sender.go
n3wscott: there is a nil aware wrapper.

We can bikeshed on the interface name, but it should not change what is in it, nothing external consumes that interface.

pkg/milestone/sender.go
n3wscott: Yes, the point is you can make your own and inject it, so it needs to be public.

pkg/milestone/sender.go
n3wscott: This way others can also make their own cloudevent impl and use this.

pkg/milestone/sender.go
n3wscott: it the break glass method

pkg/milestone/sender.go
slinkydeveloper: What about a constructor method then?

pkg/milestone/sender.go
slinkydeveloper: ?!

pkg/milestone/sender.go
n3wscott: No need to make a constructor with such a simple object, go wants this to be just an open object

pkg/milestone/sender.go
n3wscott: Renamed to `Event`

slinkydeveloper (approved):

PR #112, opened 2021-03-05
Serialize stepfn execution

slinkydeveloper:
This is an attempt to address the issues described in #106 just serializing the execution of every stepfn.

This approach should allow us to reenable other features of `testing.T`, like `Cleanup` and `Deadline` (they're not reenabled here, but we can do that in a follow up)

# Changes

- :broom: Serialize StepFn execution
- :gift: Implement requirement execution policy
- :gift: Skip asserts if setup failed


slinkydeveloper:
cc @dprotaso @n3wscott WDYT about this PR to start approaching the issues of #106?

pkg/environment/execution.go
```diff
@@ -0,0 +1,25 @@
+package environment
+
+import "knative.dev/reconciler-test/pkg/feature"
+
+// reorderSteps reorders the steps based on their timings: Setup, Requirement, Assert, Teardown
+func reorderSteps(steps []feature.Step) []feature.Step {
```
n3wscott, line 6: this looks like a micro optimization that makes it harder to reason about what `env.Test` is doing. In my opinion it makes maintaining the code harder.

pkg/environment/t.go
```diff
@@ -0,0 +1,91 @@
...
+type t struct {
+	*testing.T
+
+	failed  *atomic.Bool
+	skipped *atomic.Bool
+}
+
+var _ feature.T = (*t)(nil)
+
+func wrapT(originalT *testing.T) *t {
```
n3wscott, line 21: I do not understand the value in this.

pkg/environment/execution.go
slinkydeveloper: This is not a micro optimization, this is just making sure that the execution order is:

execute all Setup steps -> execute all Requirement steps -> execute all Assert steps -> execute all Teardown steps

pkg/environment/t.go
slinkydeveloper: This is the comment that explains it: https://github.com/knative-sandbox/reconciler-test/pull/112/files#diff-36dc7755231472dee1b65e70532635dce8e7d0edcfec67c9db7b1d8c0307998dR177

slinkydeveloper: Also from #106:

> We don't really implement Requirement as we would love to: if a Requirement fails, we don't want the test to fail, but we want to skip the asserts (don't execute them at all) and just run the teardown. More generally, we cannot really control what to do after a StepFn fails or succeds, which is preferable given our APIs and the meanings of our "phases".

This wrapped t allows us to do so

slinkydeveloper:
@n3wscott @dprotaso I've refactored again `execution.go` and added a test to `environment_test.go` to show the potential of my idea here. In short, I've implemented two policies:

* If a requirement fails, no assertions are executed, but teardown are, and the test is marked as **skipped** more than failed, with the proper cause
* If a setup fails, no requirements and assertions are executed, but teardown are, and the test is marked as **failed**

I've also added a comment about how IMO we should execute asserts and record their results, aka run them all and then check if we reached a certain compliance level.

There is some duplicated code I know, I'll remove it later, but now it's ok and it shows my point. For some reason github actions are not executing e2e tests, but you can try these on your local machine and you see they work as expected

pkg/environment/t.go
slinkydeveloper: Check https://github.com/knative-sandbox/reconciler-test/pull/112#issuecomment-792637913

pkg/environment/execution.go
slinkydeveloper: Removed this function

pkg/environment/t.go
```diff
@@ -0,0 +1,88 @@
...
+	t.failed.Store(true)
+}
+
+func (t *requirementT) Fatal(args ...interface{}) {
+	t.T.Helper()
+	t.failed.Store(true)
+	t.T.Log(args...)
+}
+
+func (t *requirementT) Fatalf(format string, args ...interface{}) {
```
n3wscott, line 52: A test will expect nothing is executed after a call to `t.Fatal/Fatalf`, this changes the go testing runtime contract.

test/e2e/environment_test.go
```diff
@@ -76,6 +76,99 @@ func TestTimingConstraints(t *testing.T) {
...
+	feat := &feature.Feature{}
+
+	counter := int32(0)
+
+	feat.Setup("setup1", appender(stringBuilder, "setup1"))
+	feat.Setup("setup2", appender(stringBuilder, "setup2"))
+	feat.Setup("setup3", appender(stringBuilder, "setup3"))
+	feat.Requirement("requirement1", appender(stringBuilder, "requirement1"))
+	feat.Requirement("requirement2", func(ctx context.Context, t feature.T) {
+		t.Fatalf("We can't fulfill this requirement")
```
n3wscott, line 99: Add a step after `t.Fatalf` that should run but will.

pkg/environment/t.go
slinkydeveloper: Is that's the case, then the very concept of our `Requirement` phase is broken and should be removed.

test/e2e/environment_test.go
slinkydeveloper: Which step? There is a already another step `requirement3` which is not executed

pkg/environment/t.go
slinkydeveloper: Ohhhh

```
   environment_test.go:100: We can't fulfill this requirement
    testing.go:1048: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
```

I may need to find a way around it...

slinkydeveloper: Skipping does the trick

test/e2e/environment_test.go
n3wscott: add `appender(stringBuilder, "requirement2")` after `t.Fatalf("We can't fulfill this requirement")` ?

n3wscott (approved):
/lgtm
/approve

PR #114, opened 2021-03-08
Added feature.Assert shortcut

slinkydeveloper:

Because when developing e2e/integration tests the user doesn't care about the compliance levels, this PR adds a `feature.Assert` which defaults to the highest level and state 

# Changes

- :gift: Add `feature.Assert` for e2e tests developers

### DIFF
--- a/pkg/feature/feature.go
+++ b/pkg/feature/feature.go
@@ -105,6 +105,12 @@ func (f *Feature) Requirement(name string, fn StepFn) {
 	})
 }
 
+// Assert is a shortcut for Stable(name).Must(name, fn),
+// useful for developing integration tests that doesn't require assertion levels.
+func (f *Feature) Assert(name string, fn StepFn) {
+	f.Stable(name).Must(name, fn)
+}
+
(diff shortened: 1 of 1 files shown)


pkg/feature/feature.go
```diff
@@ -105,6 +105,12 @@ func (f *Feature) Requirement(name string, fn StepFn) {
 	})
 }
 
+// Assert is a shortcut for Stable(name).Must(name, fn),
+// useful for developing integration tests that doesn't require assertion levels.
+func (f *Feature) Assert(name string, fn StepFn) {
```
n3wscott, line 110: I think we should get people to pick levels and spec words and not shortcut it, this defeats the purpose of defining the levels.

slinkydeveloper: Yes, but doing so in e2e tests doesn't really make sense, that's why I propose this method. In e2e tests you just want any assertion at the higher level possible

n3wscott (approved):
/lgtm
/approve

PR #116, opened 2021-03-08
Fix milestone events for new execution phases

n3wscott:


pkg/environment/magic.go
```diff
@@ -177,6 +177,11 @@ func (mr *MagicEnvironment) Prerequisite(ctx context.Context, t *testing.T, f *f
 func (mr *MagicEnvironment) Test(ctx context.Context, originalT *testing.T, f *feature.Feature) {
 	originalT.Helper() // Helper marks the calling function as a test helper function.
 
+	mr.milestones.TestStarted(f.Name, originalT.Name())
+	originalT.Cleanup(func() {
+		mr.milestones.TestFinished(f.Name, originalT.Name(), originalT.Skipped(), originalT.Failed())
```
vaikas, line 182: Seeing this called, why not change the signature to:
(name string, originalTest *testing.T)

That way if we ever need anything else from the original T we don't have to change the signature and callers?

pkg/environment/magic.go
```diff
@@ -194,12 +199,17 @@ func (mr *MagicEnvironment) Test(ctx context.Context, originalT *testing.T, f *f
...
-		var internalT *testing.T
+		var internalT feature.T
 		originalT.Run(s.T.String()+"/"+s.TestName(), func(st *testing.T) {
 			internalT = st
 			st.Helper()
 			st.Cleanup(wg.Done) // Make sure wg.Done() is always invoked, no matter what
 
+			mr.milestones.StepStarted(f.Name, s.Name, s.T.String(), s.L.String(), st.Name())
+			originalT.Cleanup(func() {
+				mr.milestones.StepFinished(f.Name, s.Name, s.T.String(), s.L.String(), st.Name(), internalT.Skipped(), internalT.Failed())
```
vaikas, line 210: And the same here:
StepFinished(featureName string, step *Step, origTest *testing.T)

vaikas (commented):
Just a thought about the signatures, just curious if they are going to get unwieldy.

pkg/environment/magic.go
n3wscott: sounds like a good idea!

pkg/environment/magic.go
n3wscott: good idea!

pkg/milestone/emitter.go
```diff
@@ -45,10 +47,12 @@ type Emitter interface {
 	Environment(env map[string]string)
 	NamespaceCreated(namespace string)
 	NamespaceDeleted(namespace string)
-	TestStarted(feature, stepName, testName string)
-	TestFinished(feature, stepName, testName string, skipped, failed bool)
-	TestSetStarted(featureSet, testName string)
-	TestSetFinished(featureSet, testName string, skipped, failed bool)
+	TestStarted(feature string, t feature.T)
+	TestFinished(feature string, t feature.T)
+	StepStarted(feature string, step *feature.Step, t feature.T)
```
vaikas, line 52: zomg one more suggest now that I look at this, sorry :)

feature string, t feature.T, step *feature.Step

Just for consistency of the API? So, featureName first, then Feature, then Step.

vaikas: Ok, I guess feature.Step is more higher level, I had it backwards. LGTM.

vaikas (approved):

test/e2e/environment_fails_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"knative.dev/reconciler-test/pkg/feature"
+)
+
+// These are example tests that fail intentionally to demonstrate feature timings working.
+
+func TestRequirementSkip(t *testing.T) {
```
slinkydeveloper, line 35: This test should not fail, it should just be skipped

Issue #40, opened 2020-11-17
Failing tests leaves namespace up 


slinkydeveloper:
This may be useful for debugging, but I guess it should be opt-in and by default the framework should just cleanup the namespace from the cluster

tayarani:
It's not just failing tests. Even successful runs leave the ns around.

slinkydeveloper:
@n3wscott how about when configuring the environment with `globalEnv.Environment` we pass `testing.T`? So then we can hook the finish to the test lifecycle:

```
globalEnvironment.Environment(t, ....)
``` 

Inside `Environment`:

```
t.Cleanup(this.Finish)
```

n3wscott:
we can add a flag in env to keep namespaces around on finish.

PR #123, opened 2021-03-10
auto manage lifecycle of env

n3wscott:
fixes https://github.com/knative-sandbox/reconciler-test/issues/40

We provide a hook to allow a user to opt into a managed environment, they are free to not use this option and call `env.Finish()` directly. 

This is a non-API breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,20 +92,20 @@ func TestFoo(t *testing.T) {
 
 	// Create an instance of an environment. The environment will be configured
 	// with any relevant configuration and settings based on the global
-	// environment settings. Additional options can be passed to Environment()
-	// if customization is required.
-	ctx, env := global.Environment(/* optional environment options */)
+	// environment settings. Using environment.Managed(t) will call env.Finish()
+	// on test completion. If this option is not used, the test should call
+	// env.Finish() to perform cleanup at the end of the test. Additional options
+	// can be passed to Environment() if customization is required.
+	ctx, env := global.Environment(environment.Managed(t) /*, optional environment options */)
 
 	// With the instance of an Environment, perform one or more calls to Test().
 	env.Test(ctx, t, FooFeature1())
 	// Note: env.Test() is blocking until the feature completes.
     env.Test(ctx, t, FooFeature2())
-
-	// Call Finish() on the Environment when finished. This will clean up any
-	// ephemeral resources created from global.Environment() and env.Test().
-	env.Finish()
 }
 ```
+
+
 
 The role of the `Test<Name>` methods is to control which features are tested on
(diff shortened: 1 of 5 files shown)


pkg/environment/magic.go
```diff
@@ -87,6 +87,15 @@ func (mr *MagicEnvironment) Finish() {
 	mr.milestones.Finished()
 }
 
+// Managed enables auto-lifecycle management of the environment. Including:
+//  - registers a t.Cleanup callback on env.Finish().
+func Managed(t feature.T) EnvOpts {
```
slinkydeveloper, line 92: I see your point of view, we might not want to tie the lifecycle of Environment to testing.t, but I wonder, since I see this less as an option and more as a "different way" of building the Environment, would it make sense to have a constructor ad hoc for it, like `globalEnv.ManagedEnvironment`? This ensures is also **more visible** and in 99% of use cases, you should use this more than `Environment`

slinkydeveloper, line 92: Not sure we should use `feature.T` or `*testing.T`, maybe let's leave `feature.T` now and in future we might use `*testing.T` if we want to.

n3wscott: I think that seems like a reasonable followup to this.

n3wscott: perhaps in the followup we can make `globalEnv.ManagedEnvironment(t *testing.T)`

slinkydeveloper: Nah let's keep `feature.T`, it's better, since we can move anyday to `*testing.T` painlessly

README.md
```diff
@@ -92,21 +92,21 @@ func TestFoo(t *testing.T) {
...
 
 	// With the instance of an Environment, perform one or more calls to Test().
 	env.Test(ctx, t, FooFeature1())
 	// Note: env.Test() is blocking until the feature completes.
-    env.Test(ctx, t, FooFeature2())
-
-	// Call Finish() on the Environment when finished. This will clean up any
-	// ephemeral resources created from global.Environment() and env.Test().
-	env.Finish()
+	env.Test(ctx, t, FooFeature2())
```
vaikas, line 104: Nit, I'd add this note to the first use of the env.Test. So, move this between L101, L102

vaikas (approved):
/hold
if you want to clean up, otherwise, remove hold.

README.md
n3wscott: Good point. Moved.

n3wscott:
/unhold

Thanks for the review @vaikas

vaikas (approved):

PR #127, opened 2021-03-11
adding helper to wait for an address for a resource.

n3wscott:
This also adds timing defaults to/from context and a way to set that on env creation.

No API breaking changes.

### DIFF
--- a/pkg/k8s/wait.go
+++ b/pkg/k8s/wait.go
@@ -19,7 +19,6 @@ package k8s
 import (
 	"context"
 	"fmt"
-
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"knative.dev/pkg/apis"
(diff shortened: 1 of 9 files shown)


pkg/k8s/wait.go
```diff
@@ -139,3 +139,27 @@ func WaitForPodRunningOrFail(ctx context.Context, t feature.T, podName string) {
 		t.Fatalf("Failed while waiting for pod %s running: %+v", podName, errors.WithStack(err))
 	}
 }
+
+// WaitForAddress waits until a resource has an address.
+func WaitForAddress(ctx context.Context, gvr schema.GroupVersionResource, name string, interval, timeout time.Duration) (*apis.URL, error) {
```
vaikas, line 144: Thanks! I do wonder if we could/might want to have a simpler version (in addition to this method) where we default interval/timeout. I reckon in most cases defaults will suffice and will then not require everybody to define the same constants over and over again :)

vaikas (approved):

pkg/k8s/wait.go
n3wscott: great idea.

pkg/k8s/wait.go
```diff
@@ -42,13 +41,42 @@ import (
...
+	if len(timing) == 1 {
+		interval = &timing[0]
+	}
+
+	di, timeout := environment.PollTimingsFromContext(ctx)
+	if interval == nil {
+		interval = &di
+	}
+
+	return *interval, timeout
```
vaikas, line 66: Just curious why interval is derefed but timeout is not?

n3wscott: interval can come from the passed slice or from the context, so I am using a pointer to help keep track

n3wscott: maybe it is complicators gloves, but this is enabling: 

```
log, err := k8s.WaitForJobTerminationMessage(ctx, client, env.Namespace(), name, time.Second, 30*time.Second)


// or just interval:

log, err := k8s.WaitForJobTerminationMessage(ctx, client, env.Namespace(), name, time.Second)

```

vaikas: Makes sense.

vaikas (approved):

PR #131, opened 2021-03-12
add variant of install from other places

vaikas:
Add ability to parse files that are non-local to the .go file that consumes them. Meant to be able to be used to test different brokers against the conformance tests.

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


n3wscott (approved):
/lgtm
/approve

PR #136, opened 2021-03-16
Enabling helper methods to merge a template with an override.

n3wscott:


pkg/manifest/templates.go
```diff
@@ -102,15 +107,25 @@ func ParseTemplates(path string, images map[string]string, cfg map[string]interf
 }
 
 // ExecuteLocalYAML will look in the callers filesystem and process the
-// templates found in files named "*.yaml" and return the f.
+// templates found in files named "*.yaml".
+// Depreciated: use ExecuteYAML(images, cfg)
```
vaikas, line 111: nit, deprecated.

pkg/manifest/templates.go
```diff
@@ -137,7 +152,73 @@ func removeComments(in string) string {
 	return regex.ReplaceAllString(in, "")
 }
 
-// OutputYAML writes out each file contents  to out after removing comments and
+func MergeYAML(files map[string]string, overlays map[string]string) ([]unstructured.Unstructured, error) {
```
vaikas, line 155: I think this could use a comment or two on what it does.

pkg/manifest/templates_test.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	results, err := manifest.MergeYAML(files, overrides)
+	if err != nil {
+		panic(err)
+	}
+
+	manifest.OutputUnstructuredAsYAML(os.Stdout, results)
```
vaikas, line 134: Could we add a test that we know we're going to need for Broker. Namely, writing the metadata.annotations field?

n3wscott: added.

pkg/manifest/templates.go
n3wscott: added a comment

pkg/manifest/templates.go
n3wscott: idk how to spell, sorry

pkg/manifest/templates.go
```diff
@@ -102,15 +107,25 @@ func ParseTemplates(path string, images map[string]string, cfg map[string]interf
...
 func ExecuteLocalYAML(images map[string]string, cfg map[string]interface{}) (map[string]string, error) {
-	pwd, _ := os.Getwd()
-	log.Println("PWD: ", pwd)
 	_, filename, _, _ := runtime.Caller(1)
 
 	return ExecuteTemplates(path.Dir(filename), "yaml", images, cfg)
 }
 
+// ExecuteYAML will look in the callers filesystem and process the
+// templates found in files named "./<f/r/o/m>/*.yaml". Relative to the caller.
```
vaikas, line 119: Why do we want it to be relative to the caller? Since these are mainly consumed by conformance tests where pieces come that are not related to caller, seems odd?

n3wscott: fair, let me think about it

slinkydeveloper: To me it seems reasonable to have it relative to the caller. Maybe we need another one where from is absolute (and not a vararg maybe?)

pkg/manifest/templates.go
```diff
@@ -102,15 +107,25 @@ func ParseTemplates(path string, images map[string]string, cfg map[string]interf
 }
 
 // ExecuteLocalYAML will look in the callers filesystem and process the
-// templates found in files named "*.yaml" and return the f.
+// templates found in files named "*.yaml".
+// Deprecated: use ExecuteYAML(images, cfg)
 func ExecuteLocalYAML(images map[string]string, cfg map[string]interface{}) (map[string]string, error) {
```
slinkydeveloper, line 112: Why this is deprecated? It seems to me an alternative of `ExecuteYAML`

pkg/manifest/templates_test.go
```diff
@@ -0,0 +1,197 @@
...
+	// Output:
```
slinkydeveloper, line 177: Is it expected that there isn't any `Foo` there?

slinkydeveloper (commented):
Is there any reason for the yaml to merge to be the "full" spec + the changes? Wouldn't it be better to just specify the difference, eg:

base:

```yaml
apiVersion: example.knative.dev/v1
kind: AAA
metadata:
  name: aaa-{{ .name }}
  namespace: {{ .namespace }}
spec:
  aaa: "{{ .aaaMsg }}"
```

Override:

```
metadata:
  annotations:
    some-custom-annotation: for all AAAs
```

Identifying what has to be overriden can be done of name basis or some key that identifies the template "id".

Also, given you use the GVK to identify which templates to merge, what happens if you need to have several yamls for the same GVK but different resource (e.g. different name/namespace)?

PR #139, opened 2021-03-16
don't use panics.

n3wscott:
Too many of these errors:

```
    --- FAIL: TestBrokerDirect/Assert/[Alpha/MUST]RabbitMQ_source_the_recorder_received_all_sent_events_within_the_time (240.00s)
panic: Fail in goroutine after TestBrokerDirect/Setup/install_recorder has completed [recovered]
	panic: Fail in goroutine after TestBrokerDirect/Setup/install_recorder has completed
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,7 +295,7 @@ func Install(message string) feature.StepFn {
 	}
 	return func(ctx context.Context, t *testing.T) {
 		if _, err := manifest.InstallLocalYaml(ctx, cfg); err != nil {
-			t.Fatal(err)
+			t.Error(err)
 		}
 	}
 }
--- a/README.md
+++ b/README.md
@@ -295,7 +295,7 @@ func Install(message string) feature.StepFn {
 	}
 	return func(ctx context.Context, t *testing.T) {
 		if _, err := manifest.InstallLocalYaml(ctx, cfg); err != nil {
-			t.Error(err)
+			t.Fatal(err)
 		}
 	}
 }
--- a/pkg/eventshub/event_info_store.go
+++ b/pkg/eventshub/event_info_store.go
@@ -158,7 +158,8 @@ func (ei *Store) Find(matchers ...EventInfoMatcher) ([]eventshub.EventInfo, even
 func (ei *Store) AssertAtLeast(min int, matchers ...EventInfoMatcher) []eventshub.EventInfo {
 	events, err := ei.waitAtLeastNMatch(allOf(matchers...), min)
 	if err != nil {
-		ei.t.Fatalf("Timeout waiting for at least %d matches.\nError: %+v", min, errors.WithStack(err))
+		ei.t.Errorf("Timeout waiting for at least %d matches.\nError: %+v", min, errors.WithStack(err))
+		return nil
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -295,7 +295,7 @@ func Install(message string) feature.StepFn {
 	}
 	return func(ctx context.Context, t *testing.T) {
 		if _, err := manifest.InstallLocalYaml(ctx, cfg); err != nil {
-			t.Fatal(err)
+			t.Error(err)
```
slinkydeveloper, line 298: This has to panic :smile:

pkg/eventshub/event_info_store.go
```diff
@@ -158,7 +158,8 @@ func (ei *Store) Find(matchers ...EventInfoMatcher) ([]eventshub.EventInfo, even
 func (ei *Store) AssertAtLeast(min int, matchers ...EventInfoMatcher) []eventshub.EventInfo {
 	events, err := ei.waitAtLeastNMatch(allOf(matchers...), min)
 	if err != nil {
-		ei.t.Fatalf("Timeout waiting for at least %d matches.\nError: %+v", min, errors.WithStack(err))
+		ei.t.Errorf("Timeout waiting for at least %d matches.\nError: %+v", min, errors.WithStack(err))
```
slinkydeveloper, line 161: This has to panic

slinkydeveloper (commented):
All the panics (which are not panics, but t.fatal) here are correct, they have to panic. If you remove the panic, it will just make harder for us to find out why they panic :smile:

README.md
n3wscott: panic here kills the entire test run

slinkydeveloper: It must kill the feature run

n3wscott:
Using fatals in the step mean the tests end early and the test run does not attempt the later asserts.

slinkydeveloper:
> Using fatals in the step mean the tests end early and the test run does not attempt the later asserts.

That is the expected behaviour... If you want to change this behaviour, you need to do it at the level of abstraction discussed in #106, not the single fatal calls here

n3wscott:
> That is the expected behaviour... If you want to change this behaviour, you need to do it at the level of abstraction discussed in #106, not the single fatal calls here

To change how this works is changing how `go test` works, I would rather make an example of not using fatals for non fatal conditions in the tests. `4 != 5` is not a fatal condition, but it is a failure; don't panic the rest of the test because a condition failed.

slinkydeveloper:
Run the "should fail" tests in this PR https://github.com/knative-sandbox/reconciler-test/pull/142

n3wscott:
> Run the "should fail" tests in this PR #142

This violates the runtime contract of `go test` is what I am trying to explain. Test authors should use fatal or panics when it truly is fatal and the test should not keep going. #142 changes the meaning of fatal for testing in go, and that is a mistake.

slinkydeveloper:
Sorry wrong button :)

README.md
n3wscott: ok this one maybe is ok

PR #148, opened 2021-03-19
add id for correlation

vaikas:
Add a field to EventInfo, Id. This is stuffed with the Event.Id so that we can actually correlate Sent / Response events
and inspect the statuscode to make sure it's what we expect.

Example usage here:
https://github.com/vaikas/eventing/commit/c9cd237e10f6e9f5f32b8a6f29a91c3247c61c99

### DIFF
--- a/pkg/test_images/eventshub/sender/sender.go
+++ b/pkg/test_images/eventshub/sender/sender.go
@@ -201,6 +201,7 @@ func Start(ctx context.Context, logs *eventshub.EventLogs) error {
 				Observer: env.SenderName,
 				Time:     time.Now(),
 				Sequence: uint64(sequence),
+				Id:       event.ID(),
 			}); err != nil {
 				return fmt.Errorf("cannot forward event info: %w", err)
 			}
@@ -212,6 +213,7 @@ func Start(ctx context.Context, logs *eventshub.EventLogs) error {
 				Observer: env.SenderName,
 				Time:     time.Now(),
 				Sequence: uint64(sequence),
+				Id:       event.ID(),
 			}
 
 			sentHeaders := make(nethttp.Header)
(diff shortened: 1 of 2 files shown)


n3wscott (approved):
/lgtm
/approve

pkg/test_images/eventshub/sender/sender.go
```diff
@@ -212,6 +213,7 @@ func Start(ctx context.Context, logs *eventshub.EventLogs) error {
 				Observer: env.SenderName,
 				Time:     time.Now(),
 				Sequence: uint64(sequence),
+				Id:       event.ID(),
```
slinkydeveloper, line 216: event might be null here, this may panic

pkg/test_images/eventshub/sender/sender.go
```diff
@@ -242,6 +244,7 @@ func Start(ctx context.Context, logs *eventshub.EventLogs) error {
 				Time:        time.Now(),
 				Sequence:    uint64(sequence),
 				StatusCode:  res.StatusCode,
+				Id:          event.ID(),
```
slinkydeveloper, line 247: Same

slinkydeveloper (commented):
You could have used just sequence to correlate :)

pkg/test_images/eventshub/sender/sender.go
slinkydeveloper: Which is the reason why this fails https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative_eventing/5108/pull-knative-eventing-reconciler-tests/1373043539906138112

slinkydeveloper:
Next time we need a change to reconciler-test, can you please try the change with your fork on eventing and then we merge here?

PR #159, opened 2021-03-31
document Store and KVStore

n3wscott:
Trying to add notes about how and why to use State.

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,7 +331,7 @@ dynamic client.
 #### Communicating State
 
 There are times when the test author would rather pass state from a `Setup` or
-`Reqirement` of a feature without a complicated parameter passing or callbacks.
+`Requirement` of a feature without a complicated parameter passing or callbacks.
 An example of this would be generating names of resources. To Address this need,
 we added `f.State`.
 
@@ -384,7 +384,7 @@ func someAssertion(ctx context.Context, t *testing.T) {
 This shows `someAssertion` can be written in a way that does not require it to
 have some complicated callback logic wrapping it but still get custom parameters
 that are dependent on the specific test run. State is useful for passing
-generated names of know keys between `Setup` and `StepFn`s. It is intended to
+generated names of known keys between `Setup` and `StepFn`s. It is intended to
 remove the need for oddly nested callbacks and globals. By default, `State` is
 scoped for the lifecycle of a `Feature` in the context of `Test`. Remember: when
 `FancyFeature` is called in the example above, its job is to compose a Feature
--- a/README.md
+++ b/README.md
@@ -342,6 +342,10 @@ State is stored into the context that is passed to each `StepFn`, implemented as
 a pointer to that feature instance's State object. Meaning, State is effectively
 global for the scope of the `env.Test` run.
 
+> Note: the default implementation in reconciler-test uses JSON marshaling to
+> save and load values to/from the store. You may provide your own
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -328,6 +328,70 @@ spec:
 Will be processed first as a go template, then applied the Environment using the
 dynamic client.
 
+#### Communicating State
+
+There are times when the test author would rather pass state from a `Setup` or
+`Reqirement` of a feature without a complicated parameter passing or callbacks.
```
vaikas, line 334: Requirement

README.md
```diff
@@ -328,6 +328,70 @@ spec:
...
+func someAssertion(ctx context.Context, t *testing.T) {
+    foo := state.GetOrFail(ctx, t, "foo")
+    // do something with foo.
+}
+```
+
+This shows `someAssertion` can be written in a way that does not require it to
+have some complicated callback logic wrapping it but still get custom parameters
+that are dependent on the specific test run. State is useful for passing
+generated names of know keys between `Setup` and `StepFn`s. It is intended to
```
vaikas, line 387: s/know keys/known keys/?

README.md
```diff
@@ -328,6 +328,70 @@ spec:
...
+
+In use,
+
+```go
+func FancyFeature(brokerName string) *feature.Feature {
+	f := feature.NewFeatureNamed("MyNewFeature")
+
+	f.Setup("Set Foo", func(ctx context.Context, t *testing.T) {
+		state.SetOrFail(ctx, t, "foo", "bar-"+random)
+    })
```
lionelvillard, line 370: indent is off

n3wscott: thanks!

README.md
```diff
@@ -328,6 +328,70 @@ spec:
...
+	})
+
+	f.Stable("An aspect of the feature").
+		Should("make some assertion", someAssertion)
+
+	return f
+}
+
+func someAssertion(ctx context.Context, t *testing.T) {
+    foo := state.GetOrFail(ctx, t, "foo")
```
lionelvillard, line 379: must be `GetStringOrFail`

README.md
```diff
@@ -328,6 +328,70 @@ spec:
...
+// Direct Store access:
+store := state.FromContext(ctx) Store
+
+// Helpers to work with Store without needing fetch Store directly.
+
+aString := state.GetStringOrFail(ctx, t, key)
+
+state.GetOrFail(ctx, t, key, value)
+
+state.SetOrFail(ctx, t, key, value)
```
lionelvillard, line 359: Maybe mentioned here that under the cover the value is marshaled and only exported fields are stored.

README.md
n3wscott: `GetStringOfFail` is a special case for known keys as strings

README.md
```diff
@@ -328,6 +328,70 @@ spec:
...
+import knative.dev/reconciler-test/pkg/state
+
+// Direct Store access:
+store := state.FromContext(ctx) Store
+
+// Helpers to work with Store without needing fetch Store directly.
+
+aString := state.GetStringOrFail(ctx, t, key)
+
+state.GetOrFail(ctx, t, key, value)
```
n3wscott, line 357: value must be a pointer.

README.md
n3wscott: AI: I need to followup to see if this is the right syntax here.

PR #163, opened 2021-04-05
[EventsHub] Refactor sender to use a generator

n3wscott:
I want to be able to configure the sender to use an input yaml file, but I need to stage a bit of a refactor to get the logic seperated a little. 1/2

### DIFF
--- a/pkg/test_images/eventshub/sender/sender.go
+++ b/pkg/test_images/eventshub/sender/sender.go
@@ -20,27 +20,25 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"flag"
 	"fmt"
 	"io/ioutil"
 	nethttp "net/http"
 	"strconv"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/wait"
-
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	"github.com/cloudevents/sdk-go/v2/binding"
 	cehttp "github.com/cloudevents/sdk-go/v2/protocol/http"
 	"github.com/kelseyhightower/envconfig"
 	"go.opencensus.io/plugin/ochttp"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"knative.dev/pkg/logging"
 	"knative.dev/pkg/tracing/propagation/tracecontextb3"
 
 	"knative.dev/reconciler-test/pkg/test_images/eventshub"
 )
 
-type envConfig struct {
+type generator struct {
 	SenderName string `envconfig:"POD_NAME" default:"sender-default" required:"true"`
 
 	// Sink url for the message destination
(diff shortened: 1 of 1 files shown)


pkg/test_images/eventshub/sender/sender.go
```diff
@@ -87,21 +85,27 @@ type envConfig struct {
...
+		return err
+	}
 
 	logging.FromContext(ctx).Infof("Sender environment configuration: %+v", env)
 
-	if env.InputEvent == "" && env.InputBody == "" && len(env.InputHeaders) == 0 {
-		return fmt.Errorf("input values not provided")
-	}
-
-	flag.Parse()
```
vaikas, line 104: There was no need for this?

pkg/test_images/eventshub/sender/sender.go
```diff
@@ -146,152 +141,192 @@ func Start(ctx context.Context, logs *eventshub.EventLogs) error {
...
+		case <-ticker.C:
+			// Keep looping.
+		case <-ctx.Done():
+			logging.FromContext(ctx).Infof("Canceled sending messages because context was closed")
+			return nil
 		}
+	}
+}
+
+func (g *generator) sentInfo(event *cloudevents.Event, req *nethttp.Request, err error) eventshub.EventInfo {
```
vaikas, line 190: just curious since you don't use the event for anything except the ID, if it would be cleaner to just pass in the eventID here instead of the whole event?

pkg/test_images/eventshub/sender/sender.go
```diff
@@ -146,152 +141,192 @@ func Start(ctx context.Context, logs *eventshub.EventLogs) error {
...
+	}
+	sentEventInfo.HTTPHeaders = sentHeaders
+
+	if g.InputBody != "" {
+		sentEventInfo.Body = []byte(g.InputBody)
+	}
+	return sentEventInfo
+}
+
+func (g *generator) responseInfo(res *nethttp.Response, event *cloudevents.Event) eventshub.EventInfo {
```
vaikas, line 230: same here about just passing the ID.

pkg/test_images/eventshub/sender/sender.go
n3wscott: nope, no flags are used

pkg/test_images/eventshub/sender/sender.go
n3wscott: It does log the entire event: 

```
	sentEventInfo := eventshub.EventInfo{
...
		Event:    event,
...
	}
```

pkg/test_images/eventshub/sender/sender.go
n3wscott: I think logging the event might be a waste of time, if we have the headers and body too...

vaikas (approved):
/lgtm

PR #172, opened 2021-04-12
move eventshub to single dir

n3wscott:


pkg/eventshub/assert/event_info_matchers.go
```diff
@@ -23,22 +23,21 @@ import (
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	cetest "github.com/cloudevents/sdk-go/v2/test"
 
-	pkgeventshub "knative.dev/reconciler-test/pkg/eventshub"
-	"knative.dev/reconciler-test/pkg/test_images/eventshub"
+	"knative.dev/reconciler-test/pkg/eventshub"
 )
 
 // EventInfo is a re-export of eventshub.EventInfo for convenience
 type EventInfo = eventshub.EventInfo
```
vaikas, line 30: Just curious if this is still necessary after the moves?

pkg/k8s/wait.go
```diff
@@ -201,3 +200,59 @@ func WaitForAddress(ctx context.Context, gvr schema.GroupVersionResource, name s
 	})
 	return addr, err
 }
+
+// WaitForServiceHasAtLeastOneEndpoint polls the status of the specified Service
```
vaikas, line 204: This comment needs to be updated.

Also, curious if we should change it to not look for exact match but for >= endpoints?

pkg/k8s/wait.go
```diff
@@ -201,3 +200,59 @@ func WaitForAddress(ctx context.Context, gvr schema.GroupVersionResource, name s
...
+	endpointsService := client.CoreV1().Endpoints(svcNamespace)
+
+	interval, timeout := PollTimings(ctx, nil)
+	return wait.PollImmediate(interval, timeout, func() (bool, error) {
+		endpoint, err := endpointsService.Get(ctx, svcName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+
+		return countEndpointsNum(endpoint) == numOfEndpoints, nil
```
vaikas, line 216: so here >= numOfEndpoints?

pkg/eventshub/assert/event_info_matchers.go
n3wscott: sounds good.

pkg/k8s/wait.go
n3wscott: updated!

vaikas (approved):
/lgtm

PR #175, opened 2021-04-14
Adding prober reject settings and examples.

n3wscott:


pkg/eventshub/prober.go
```diff
@@ -136,6 +153,31 @@ func (p *EventProber) SenderDone(prefix string) feature.StepFn {
...
+// ReceiverDone will poll until the receiver has received all expected events.
+func (p *EventProber) ReceiverDone(from, to string) feature.StepFn {
+	return func(ctx context.Context, t feature.T) {
+		interval, timeout := environment.PollTimingsFromContext(ctx)
+		err := wait.PollImmediate(interval, timeout, func() (bool, error) {
+			sent := p.SentBy(ctx, from)
+			fmt.Println(from, "has sent", len(sent))
+
+			received := p.ReceivedBy(ctx, to)
+			fmt.Println(to, "has received", len(received))
```
vaikas, line 165: Sure would be nice to plumb the logger here so we'd at least get some timing information. 🤔

pkg/eventshub/prober.go
```diff
@@ -221,25 +285,40 @@ func (p *EventProber) SenderEventsFromSVC(svcName, path string) feature.StepFn {
...
+// Warning: only call once.
 func (p *EventProber) SenderFullEvents(count int) {
-	for i := 0; i < count; i++ {
+	event := cetest.FullEvent()
+	if count == 1 {
 		id := uuid.New().String()
-		event := cetest.FullEvent()
 		event.SetID(id)
-		p.ids = append(p.ids, id)
+		p.ids = []string{id}
```
vaikas, line 294: This is kind of wonky that you have to special case sending 1 event here and in SenderMinEvents.

n3wscott: it is a limitation of eventshub :idk:, I assumed you could chain these together originally, and that is not true.

pkg/eventshub/prober.go
n3wscott: that was debug, I can remove now or in a followup. We should have a better story about the logger.

pkg/eventshub/prober.go
vaikas: Can we create an issue for it?

vaikas (approved):
/lgtm

PR #177, opened 2021-04-14
add kafka to downstreams

n3wscott:


benmoss:
I think this needs to be rebased so that it has `feature.FromContext`. Weird stuff with the downstream tests.

PR #178, opened 2021-04-19
Implement step planning event

n3wscott:
This change enables consumers of the event milestone stream plan ahead for what results are going to come for a single feature run. This could be used to implement something like a progress bar, etc.


vaikas (approved):
/lgtm

PR #182, opened 2021-04-20
Fail on feature.Require errors.

### LINKED ISSUE #181: feature requirements should fail the feature if the step fails, not skip
example:

``` 
   channel_test.go:51: Skipping feature 'Knative Channel Specification - Data Plane/Delivery Spec 0' assertions because requirement "wait until done" failed
```

results in a skipped step and a `PASS`, but the test did not run or pass. These should `FAIL`.

n3wscott:
fixes #181 

- **Requirement** is used to validate the cluster, environment, or anything
  else. Think of this as a preflight validation for the assertions. This can be
  used as a fast-fail for a feature test before running assert phase steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,15 +128,16 @@ ordering is required it is recommended to break that into an independent feature
 that is tested on an environment in order required.
 
 Features have 4 phases (timing) on which steps can be composed: Setup,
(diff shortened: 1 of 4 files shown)


vaikas (approved):
/lgtm

PR #185, opened 2021-04-21
Delete feature resources

n3wscott:
Clean up yo mess

### DIFF
--- a/pkg/feature/feature.go
+++ b/pkg/feature/feature.go
@@ -19,14 +19,14 @@ package feature
 import (
 	"context"
 	"fmt"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"knative.dev/pkg/apis"
-	"knative.dev/pkg/injection/clients/dynamicclient"
 	"runtime"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/feature/feature.go
```diff
@@ -93,10 +97,31 @@ func (f *Feature) Reference(ref ...corev1.ObjectReference) {
 	f.refs = append(f.refs, ref...)
 }
 
+// References returns all known resources to the Feature registered via
+// `Reference`.
 func (f *Feature) References() []corev1.ObjectReference {
 	return f.refs
 }
 
+// DeleteResourcesFn delete all known resources to the Feature registered
```
vaikas, line 106: I was going to add this myself after doing it in the eventing, but got me thinking if we should instead of this have a DoForResources( or something like that) instead of this DeleteResources. I'm fine with this going in as is but I think in the future I wonder if we should be able to control for example conditionally deleting the resources iff the feature failed for example, or something like that. Just a thought experiment? If nothing else, I do wonder if we should implement DeleteResources on top of DoForResources so we can implement more neatness later?

n3wscott: want to land this and refactor before we downstream it, or want me to take a crack at `DoForResources`?

vaikas (approved):
/lgtm

PR #188, opened 2021-04-23
#180 add flag to allow filtering of features while running tests

### LINKED ISSUE #180: Make it possible to run tests against a single feature.
As a test writer (working on creating a new test), and or vendor debugging a failing conformance test, it would be desireable to run a single feature only. For example, while working on:
https://github.com/knative/eventing/pull/5282

It was possible for me to run only the TestBrokerConformance tests, but even then it would run a whole bunch of tests that I did not want (well, need) to run.

I think what I'd like to have is something that when I today run something like this:
```
vaikas@villes-mbp reconciler-test % BROKER_CLASS=RabbitMQBroker BROKER_TEMPLATES=/Users/vaikas/projects/go/src/knative.dev/eventing-rabbitmq/test/conformance/testdata/with-operator SYSTEM_NAMESPACE=knative-eventing go test -count=1 -v -tags=e2e -run TestBrokerConformance ./test/rekt/...
```

I'd like to add a new Flag to it to restrict it to only:
https://github.com/knative/eventing/blob/main/test/rekt/features/broker/control_plane.go#L60

You can do this today hackily by creating a new top level test function but it's not great for 'vendor debugging' use case in particular.

Furthermore, if we change this function to create a new Feature for each table test entry:
https://github.com/knative/eventing/blob/main/test/rekt/features/broker/control_plane.go#L60

Then targeting a particular feature should be possible by adding a new flag, call it featureMatch

So if instead of the above, I'd do:
```
vaikas@villes-mbp reconciler-test % BROKER_CLASS=RabbitMQBroker BROKER_TEMPLATES=/Users/vaikas/projects/go/src/knative.dev/eventing-rabbitmq/test/conformance/testdata/with-operator SYSTEM_NAMESPACE=knative-eventing go test -count=1 -v -tags=e2e -run TestBrokerConformance ./test/rekt/... --featureMatch="Event Routing Spec"

```

and it would only run those table tests, that would be real nice. However, if as discussed above, we change each table test to return a Feature with a name, one could see the name of the feature becoming a prefix with something like:
"Event Routing Spec - Two Triggers..."

And for cases like that, it might be useful to have a pattern match but that's for extra credit I reckon.

xtreme-sameer-vohra:
# Changes
- :gift: Add --feature flag for filtering
    - if not set, all features are run
    - if set, the features are matched using the given regexp and the
      feature is skipped if it doesn't match. Rather than use the built-in
      skip, which skips the remainder of the test, none of the feature
      phases are executed.
    - eg. `go test -v -tags=e2e ./test/example/ --feature ^Noop$`

fixes #180

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,6 +361,7 @@ disabled using the following flags:
 | --requirement.shouldnot | Boolean | Enable/Disable running Should Not level features. |
 | --requirement.may       | Boolean | Enable/Disable running May level features.        |
 | --requirement.all       | Boolean | Enable/Disable running All level features.        |
+| --feature               | RegExp  | Specify features to run.                          |
 
 They can be used in combination, to run only Beta state features:
 
@@ -378,4 +379,10 @@ And normal go filters work on the go test entry point names.
 
 ```shell
 go test -v -count=1 -tags=e2e ./test/... --feature.any=false --feature.beta -run TestKoPublish
+```
+
+Run all instances of Noop feature, including instances in a Feature Set
+
+```shell
+go test -v -count=1 -tags=e2e ./test/... --feature=Noop
 ```
--- a/pkg/environment/flags.go
+++ b/pkg/environment/flags.go
@@ -28,6 +28,7 @@ import (
 var (
 	s = new(feature.States)
 	l = new(feature.Levels)
+	f = new(string)
 )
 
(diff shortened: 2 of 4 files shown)


n3wscott (approved):
awesome!!

PR #196, opened 2021-05-05
Use go 1.16 embed for templates

### LINKED ISSUE #193: Can we leverage the embed fs from go 1.16 to reduce the disk reads?
The trademark committee has a long term goal of a binary distribution for running conformance tests. The exact method is not yet described, but one thing that is hard at the moment is the reconciler-test framework does a lot of filesystem inspection when running, meaning it really does not work well if it is run outside the root of the project you have crafted the tests for. 

There is a new feature in go 1.16 around embedding files into the go binary into an in-memory filesystem: https://golang.org/doc/go1.16#library-embed

We currently have a lot of code that makes it easy to read and parse files that are local to the go file https://github.com/knative-sandbox/reconciler-test/blob/main/test/example/config/echo/echo.go#L39 But this is all fragile if we move the filesystem around, and becomes problematic when vendoring and impossible for building that binary distro of conformance. 

It would be interesting to investigate if we could switch to using the embedded fs to access the local yaml vs reading it off the disk as the test in being run allowing us to take a step in the "build a binary" direction we will need to get to eventually. 

An alternative to consider would be urls instead of file system reads but then local changes become hard, and running that binary in an isolated env will become very hard.

benmoss:
# Changes

- :gift:  YAML helpers now take an [fs.FS](https://golang.org/pkg/io/fs/#FS), allowing us to use the new [embed package](https://golang.org/pkg/embed/) for bundling yaml into test binaries

/kind enhancement

Fixes #193

### DIFF
--- a/test/example/config/echo/echo.go
+++ b/test/example/config/echo/echo.go
@@ -18,14 +18,18 @@ package echo
 
 import (
 	"context"
+	"embed"
 
 	"knative.dev/reconciler-test/pkg/environment"
 	"knative.dev/reconciler-test/pkg/feature"
 	"knative.dev/reconciler-test/pkg/manifest"
 )
 
+//go:embed *.yaml
+var templates embed.FS
+
 func init() {
-	environment.RegisterPackage(manifest.ImagesLocalYaml()...)
+	environment.RegisterPackage(manifest.ImagesFromFS(templates)...)
 }
 
 // Output is the base output we can expect from a echo job.
@@ -36,7 +40,7 @@ type Output struct {
 
 func Install(name, message string) feature.StepFn {
 	return func(ctx context.Context, t feature.T) {
-		if _, err := manifest.InstallLocalYaml(ctx, map[string]interface{}{
+		if _, err := manifest.InstallYamlFS(ctx, templates, map[string]interface{}{
 			"name":    name,
 			"message": message,
 		}); err != nil {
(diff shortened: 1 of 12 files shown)


test/example/config/echo/echo.go
```diff
@@ -18,14 +18,18 @@ package echo
...
 	"knative.dev/reconciler-test/pkg/feature"
 	"knative.dev/reconciler-test/pkg/manifest"
 )
 
+//go:embed *.yaml
+var templates embed.FS
+
 func init() {
-	environment.RegisterPackage(manifest.ImagesLocalYaml()...)
+	environment.RegisterPackage(manifest.Images(templates)...)
```
benmoss, line 32: Not sure about this rename, `ImagesLocalYaml` seems wrong now, but `ImagesYaml` seems weird too

n3wscott: `.Images` seems fine to me

benmoss:
Not sure if we need to bump all the downstreams first, this is a breaking change

benmoss:
https://github.com/knative-sandbox/reconciler-test/pull/195 needs to merge

n3wscott:
I think for the action to update from #195, you need to rebase this PR

n3wscott:
We should get this in first thing after 0.23 is released.

n3wscott (approved):
I am really excited about this change, I can't wait to see it in action in eventing.

PR #197, opened 2021-05-07
Add Kafka projects to downstream tests

pierDipi:
`eventing-kafka` and `eventing-kafka-broker` are using rekt too!


n3wscott (approved):
/lgtm

PR #202, opened 2021-05-20
Sender now support sub-second period and delay

lionelvillard:
# Changes

- :gift: DELAY and PERIOD now supports Golang duration format. If the last unit is not specified, add 's' to the duration string.
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```

### DIFF
--- a/pkg/eventshub/sender/sender.go
+++ b/pkg/eventshub/sender/sender.go
@@ -26,6 +26,7 @@ import (
 	nethttp "net/http"
 	"strconv"
 	"time"
+	"unicode"
 
 	conformanceevent "github.com/cloudevents/conformance/pkg/event"
 	conformancehttp "github.com/cloudevents/conformance/pkg/http"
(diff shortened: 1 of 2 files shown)


benmoss:
Why don't we just allow a time.Duration string?

lionelvillard:
That's fine by me. We can even reuse DELAY and PERIOD and make `s` the default unit, for backward-compatible reason. WDYT?

benmoss:
That sounds perfect, I was thinking we'd have to introduce a new field for compatibility, but you're right we can just default to seconds if not specified

lionelvillard:
@benmoss can you TAL?

pkg/eventshub/sender/sender.go
```diff
@@ -117,8 +118,17 @@ func Start(ctx context.Context, logs *eventshub.EventLogs) error {
 
 	logging.FromContext(ctx).Infof("Sender environment configuration: %+v", env)
 
-	period := time.Duration(env.Period) * time.Second
-	delay := time.Duration(env.Delay) * time.Second
+	p, err := time.ParseDuration(durationWithUnit(env.Period))
+	if err != nil {
+		return err
+	}
+	period := time.Duration(p)
```
benmoss, line 125: Aren't `p` and `d` here already Durations?

pkg/eventshub/sender/sender.go
```diff
@@ -117,8 +118,17 @@ func Start(ctx context.Context, logs *eventshub.EventLogs) error {
...
+
+	d, err := time.ParseDuration(durationWithUnit(env.Delay))
+	if err != nil {
+		return err
+	}
+	delay := time.Duration(d)
```
benmoss, line 131:
```suggestion
	period, err := time.ParseDuration(durationWithUnit(env.Period))
	if err != nil {
		return err
	}

	delay, err := time.ParseDuration(durationWithUnit(env.Delay))
	if err != nil {
		return err
	}
```

lionelvillard: duh! Thanks!

PR #205, opened 2021-05-25
Fix redundant name in Assert helper

benmoss:
Tests are currently named things like `events_have_passed_through_transform_service_events_have_passed_through_transform_service` because name is captured twice


n3wscott (approved):
/lgtm
/approve

PR #209, opened 2021-06-01
Fail from ReceiverDone/SenderDone

benmoss:
Right now ReceiverDone and SenderDone don't fail, they erroneously call `Failed()` which just returns a bool.

### DIFF
--- a/pkg/eventshub/prober.go
+++ b/pkg/eventshub/prober.go
@@ -148,7 +148,7 @@ func (p *EventProber) SenderDone(prefix string) feature.StepFn {
 			return false, nil
 		})
 		if err != nil {
-			t.Failed()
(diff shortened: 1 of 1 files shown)


vaikas:
/hold
/lgtm
/approve
Just adding a hold since there's a failing test, from a quick look does not seem related. If you can confirm it's not related, just remove the hold at your will.

benmoss:
I wasn't able to reproduce it locally, but the downstream tests are a bit complex in how the setup works so I'm not 100% willing to rule it out. Is it possible to somehow retrigger that test?

benmoss:
Ah I can do it from my fork, let me see

benmoss:
Good/bad news! It fails even without my changes! https://github.com/benmoss/reconciler-test/pull/3/checks?check_run_id=2730370572

/hold cancel

PR #212, opened 2021-06-03
Fix ExecuteTemplates/ExecuteYAML to work with single files again

benmoss:
Need to still figure out how to make this work with the FS apis, but that is a problem for another day.


n3wscott (approved):
/lgtm
/approve

PR #217, opened 2021-06-21
Skip emitting events if they are eventshub events

benmoss:
With APIServerSources it's possible to get our own event back as a cloudevent, which causes a loop.


aliok (approved):
/lgtm
/approve

Thanks for fixing this!

@benmoss FYI, I wanted to write some test for this, but I wasn't very successful. 

1. Can't pass multiple events with `eventshub.InputEvent()`
2. Not sure how to do a negation like "assert that the event with ID abcdef doesn't exist in event store"

See https://gist.github.com/aliok/22cfa092331b31ace8a10f3ae153135d

Anyway, I tested the changes in the PR manually and they work fine!

PR #218, opened 2021-06-22
Fix eventshub race

benmoss:
# Changes

- :bug: Fix a race where we can miss events that are emitted before our event listener is registered
- :broom: Add more helpful logs with timestamps and resolved names

/kind bug


matzew (approved):
/lgtm
/approve

PR #233, opened 2021-08-11
Use namespaced RBAC for eventshub

mgencur:
# Changes

- Create ServiceAccount, Role, RoleBinding for eventshub with the same names as the namespace
- Extract the creation into a separate Install function so that we can later call this conditionally when running tests in pre-created namespace (this feature is TBD but I want to work on it soon) 

/kind cleanup

Fixes #191

**Release Note**

```release-note

```

**Docs**

```docs

```

PR #234, opened 2021-08-12
Allow for tests to run in pre-created NS

### LINKED ISSUE #192: Allow for tests to run in pre-created NS
**Problem:**

In some environments it won't be possible for the testing infrastructure to easily create new namespaces at will, so it would be nice if there was a way to run the e2e tests all within a single, pre-created, namespace.

A similar issue was created for knative/eventing a long time ago: https://github.com/knative/eventing/issues/2776. [This comment](https://github.com/knative/eventing/issues/2776#issuecomment-829088623) includes possible solutions. I'm listing them below for completeness:
* Having namespace names generated based on the test name (without random suffix) - this was problematic because tests often have sub-tests and these subtests would share the same namespace unless the individual objects have random suffixes.
* Having just one common namespace and generate every object name for the tests with random suffixes so that all objects are unique even if they exist in the same namespace. This is a lot of work, and requires big changes in the test suite. Another problem is that some individual tests might require label knative-eventing-injection": "enabled" and some require "disabled" which makes it hard to run those in parallel in the same namespace.
* Have a numeric suffix in namespaces names.
This solution is used in knative/client:
https://github.com/knative/client/blob/main/lib/test/integration.go#L47
It is pretty simple, each test has a unique namespace but the name of the namespace is predictable. Each test gets a namespace with a numeric suffix that is increasing. The suffix is a simple number so it's possible to pre-create them, apply some additional logic or/and create additional objects in them. In a specific environment, this can be done by a user with increased privileges (admin). The limitation is that individual tests should not require specific settings that would have to be done by admin user because it is not known in advance which NS the test will run in.

**Persona:**
Contributors/Devs

Exit Criteria
Ability to run e2e tests against a pre-created cluster and namespaces.

mgencur:
# Changes

This builds upon https://github.com/knative-sandbox/reconciler-test/pull/233 . I'm including the commits from that PR here as well and will remove them later when/if the other PR is merged

- Define -reusenamespace flag
- If the flag is used the test namespace is not created but re-used. The name of the namespace is typically "rekt" followed by a numeric index starting from 0. ( we used similar approach in knative/eventing). The index is increased for each new test.
- If the flag is used then also the "eventshub" installation skips installing RBAC and rather expects the RBAC (ServiceAccount,Role, RoleBinding) to already be created. This will allow for running REKT tests by a user with reduced privileges (project admin)

I tried running the rekt tests from knative/eventing with pre-created namespaces and this flag by a project admin user. The result is:
```
DONE 747 tests, 8 skipped, 27 failures in 1225.891s
```
The failures are related to several individual tests which require clusterrole creation. These tests will need to be excluded from the execution. 

/kind enhancement

Fixes #192

**Release Note**

```release-note

```

**Docs**

```docs

```


pierDipi (approved):
/lgtm

pierDipi:
@mgencur are you still interested in this PR?

mgencur:
@pierDipi No. Let me close it. If we need it again I can create a new one but it's unlikely.

PR #241, opened 2021-09-14
import release 0.26 dependencies

ZhiminXiang:


rhuss:
@ZhiminXiang some tests are failing, but just curious: Why is the dependency not automatically updated by knobots ?

ZhiminXiang:
> @ZhiminXiang some tests are failing, but just curious: Why is the dependency not automatically updated by knobots ?

not sure. I tried to kick out the automatic update from knobots, but could not successfully make it work.

ZhiminXiang:
The tests are now good. Ready for review.

PR #260, opened 2021-11-11
Retry on webhook EOF errors

pierDipi:
This patch adds a workaround for knative/pkg#1509.

In the existing eventing testing library we have a similar workaround:
https://github.com/knative/eventing/blob/a8a30634681b2f0b23e5218348d9ae9a42cc00ff/test/lib/creation.go#L55-L72

# Changes

- Retry on webhook EOF errors

/kind bug

Fixes https://github.com/knative/eventing/issues/5892

**Release Note**

```release-note
Retry on webhook EOF errors.
```

**Docs**

```docs
None
```

### DIFF
--- a/pkg/manifest/installer.go
+++ b/pkg/manifest/installer.go
@@ -138,6 +138,5 @@ func isWebhookError(err error) bool {
 	// Internal error occurred: failed calling webhook "defaulting.webhook.kafka.eventing.knative.dev": Post "https://kafka-webhook-eventing.knative-eventing.svc:443/defaulting?timeout=2s": EOF
 	return strings.Contains(str, "webhook") &&
 		strings.Contains(str, "https") &&
-		strings.Contains(str, "knative.dev") &&
 		strings.Contains(str, "EOF")
 }
--- a/pkg/manifest/installer.go
+++ b/pkg/manifest/installer.go
@@ -26,7 +26,9 @@ import (
 	"runtime"
 	"strings"
 
+	"k8s.io/client-go/util/retry"
 	"knative.dev/pkg/injection/clients/dynamicclient"
+
 	"knative.dev/reconciler-test/pkg/environment"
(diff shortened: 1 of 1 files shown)


pkg/manifest/installer.go
```diff
@@ -121,3 +130,14 @@ func ImagesFromFS(fsys fs.FS) []string {
...
 }
+
+// isWebhookError is a workaround for https://github.com/knative/pkg/issues/1509.
+func isWebhookError(err error) bool {
+	str := err.Error()
+	// Example error:
+	// Internal error occurred: failed calling webhook "defaulting.webhook.kafka.eventing.knative.dev": Post "https://kafka-webhook-eventing.knative-eventing.svc:443/defaulting?timeout=2s": EOF
+	return strings.Contains(str, "webhook") &&
+		strings.Contains(str, "https") &&
+		strings.Contains(str, "knative.dev") &&
```
odacremolbap, line 141: I know this is a workaround, and hence temporary, but maybe `knative.dev` could be removed to avoid too much coupling.

I would ask for parametrizing the contains condition but looking at the existing code it doesn't sound trivial.

odacremolbap (commented):
Code wise+1.
Unsure about assuming `eventing` if this retry workaround is potentially helpful for other webhooks.

pkg/manifest/installer.go
pierDipi: Yes, I want that it covers all other webhooks (the example is a Kafka webhook)

Sounds good to remove `knative.dev`

```suggestion
```

pierDipi: > I would ask for parametrizing the contains condition but looking at the existing code it doesn't sound trivial.

Yes, it breaks every API and assumption, so it's good for now.

odacremolbap (approved):

antoineco (approved):
Hacky! But that works :)

PR #263, opened 2021-12-02
Use foreground propagation when deleting resources

pierDipi:
Foreground propagation has a better consistency model
to deal with especially for tests where we want to delete
a resource and then recreate it.

# Changes

- Use foreground propagation when deleting resources

/kind enhancement

**Release Note**

```release-note
Use foreground propagation when deleting resources
```

**Docs**

```docs
None
```


matzew (approved):
/lgtm
/approve

PR #269, opened 2021-12-23
:bug: Ensure the `environment.InitFlags()` is called to avoid confusion.

### LINKED ISSUE #268: Running tests without calling `environment.InitFlags()` func leads to failed test being skipped instead of failure
The `environment.InitFlags()` is required to be executed. Without it, the flags are not set properly. Their values are outside of possible known values - `s == 0` and `l == 0`.

https://github.com/knative-sandbox/reconciler-test/blob/816f2192fec9b6363c2b9ed79875aa54d472eede/pkg/environment/flags.go#L29-L30

This leads to nasty behavior - the `shouldFail(s *feature.Step) bool` func returns false, and instead of failures, the tests are skipped. This behavior is hard to understand.

https://github.com/knative-sandbox/reconciler-test/blob/816f2192fec9b6363c2b9ed79875aa54d472eede/pkg/environment/magic.go#L306-L309

I think, we should amend the test framework to ensure flags have valid values. Either by forcing people to call `environment.InitFlags()`, or setting flag values with reasonable defaults.

cardil:
# Changes

- :bug: Ensure the flags are valid

/kind bug

Fixes #268

**Release Note**

```release-note
- :bug: Ensure the flags are valid
```


pkg/feature/level.go
```diff
@@ -71,6 +71,16 @@ func (l Levels) String() string {
...
+func (l Levels) Valid() bool {
+	values := []Levels{Must, MustNot, Should, ShouldNot, May, All}
+	for _, value := range values {
+		if l == value {
+			return true
+		}
+	}
+	return false
+}
+
```
dprotaso, line 83: This isn't ideal since the flags/levels are bit masks.

I'm not really a maintainer on this package - I'm surprised to see globals `l`, `s` etc. given what we've learned from `knative.dev/pkg/test` horrors

test/example/main_test.go
```diff
@@ -38,13 +38,11 @@ import (
 // config as well as the parsing level and state flags.
 var global environment.GlobalEnvironment
 
-func init() {
+// TestMain is the first entry point for `go test`.
+func TestMain(m *testing.M) {
 	// environment.InitFlags registers state and level filter flags.
 	environment.InitFlags(flag.CommandLine)
```
dprotaso, line 44: I'm surprised `InitFlags` isn't instanced - ie. we use `global.InitFlags`

pkg/feature/level.go
cardil: Oh, right. I'll rewrite that to work for all bit mask combinations.

I don't get your second sentence. My feelings towards those single character global variables in this package are negative as well. But, it wasn't me who introduced them. I could rename them, but maybe that should be done in separate PR?

test/example/main_test.go
dprotaso: `InitFlags` is initializing package global variables in reconciler-test 

When it could simply just modify values on the declared `global` environment  in the example. At least those aren't exposed publicly so maybe that's ok. :shrug:

cardil: Yes. I agree such refactor would be beneficial. Mind opening an issue to track it?

dprotaso:
there's a conflict with HEAD

/lgtm
/approve

cardil:
@dprotaso Could you slap the PR again?

pkg/feature/level_test.go
```diff
@@ -0,0 +1,114 @@
...
+	levels := make([]feature.Levels, len(subsets))
+	for i, subset := range subsets {
+		val := 0
+		for _, level := range subset {
+			val |= int(level)
+		}
+		levels[i] = feature.Levels(val)
+	}
+	return levels
+}
```
pierDipi, line 95: This is complex and hard to read code, what's the value of running all combinations when testing the boundaries is just enough?

pkg/feature/states_test.go
```diff
@@ -0,0 +1,112 @@
...
+	levels := make([]feature.States, len(subsets))
+	for i, subset := range subsets {
+		val := 0
+		for _, state := range subset {
+			val |= int(state)
+		}
+		levels[i] = feature.States(val)
+	}
+	return levels
+}
```
pierDipi, line 93: Same here, very hard to read and follow

pkg/feature/level.go
```diff
@@ -76,6 +79,16 @@ func (l Levels) MarshalJSON() ([]byte, error) {
...
 
+// Valid checks if given level is valid (not empty combination of Levels values).
+func (l Levels) Valid() bool {
+	val := uint8(l)
+	if val == 0 {
+		return false
+	}
+	val &^= uint8(All)
+	return val == 0
+}
```
pierDipi, line 90: I'd like to reduce bit manipulation as much as possible, a simple:
```

	levels := sets.NewInt32(
		int32(All),
		int32(Must),
		int32(MustNot),
		int32(Should),
		int32(ShouldNot),
		int32(May),
	)
	
	return levels.Has(int32(l))
```

is much more readable and can go a long way

Same comment for Valid for States

pkg/feature/level_test.go
cardil: This code computes the all bitwise combinations of all levels. For example: `MUST|MUST_NOT|SHOULD` or `MAY|MUST` or just `MUST`.

How about I add such doc to the test? WDYT?

cardil: I think this is solved in 7bd94c50adc57fe32382d96e41551379071d9154

pkg/feature/states_test.go
cardil: I think this is solved in 7bd94c50adc57fe32382d96e41551379071d9154

pkg/feature/level.go
cardil: It's not the same. Now, you could use bitwise or to select multiple levels or states.

pierDipi: Isn't this an input from command line? how do you do bitwise from command line?

That suggests that the type is wrong and shouldn't be a single value but rather an array

cardil: No. You could invoke any number of those levels and features. See the code:

https://github.com/knative-sandbox/reconciler-test/blob/0979798b59c810bcfc4cfa4f63f75bb8f5711020/pkg/environment/flags.go#L41-L56

Just call the tests with those flags or states:

```
-feature.alpha -feature.beta -requirement.must -requirement.should -requirement.may
```

pierDipi: Still, this suggests that an array to represents Levels and States is better

cardil: That's not the point of this PR, isn't?

PR #277, opened 2022-01-13
Improve logging for features and their state

pierDipi:
This would help debug test failures.

### DIFF
--- a/pkg/feature/feature.go
+++ b/pkg/feature/feature.go
@@ -18,6 +18,7 @@ package feature
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"runtime"
 	"strings"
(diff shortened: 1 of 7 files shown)


pkg/feature/feature.go
```diff
@@ -41,6 +42,29 @@ type Feature struct {
...
+	}{
+		Name:  f.Name,
+		Steps: f.Steps,
+		State: f.State,
+		Refs:  f.refs,
+	}
+	return json.MarshalIndent(in, "", " ")
+}
+
+func (f Feature) Logf(log func(args ...interface{})) {
```
devguyio, line 60: - The name was very confusing for me, I thought it'd be a `Formatted log of its args` . This looks more like "dump feature state using log function". Wdyt? maybe `f.DumpWith(T.Log)` or something like that
- NIT: method doc?

devguyio (commented):
nice idea! left a comment

pkg/feature/feature.go
pierDipi: Done

devguyio (approved):
/lgtm

PR #292, opened 2022-01-27
Wait pod running prints non-ready pod logs

pierDipi:
Logging pod logs of unhealthy pods helps debug test issues.

### DIFF
--- a/pkg/k8s/wait.go
+++ b/pkg/k8s/wait.go
@@ -250,11 +250,20 @@ func WaitForPodRunningOrFail(ctx context.Context, t feature.T, podName string) {
 		isRunning := podRunning(p)
 
 		if !isRunning {
-			t.Log("Pod %s/%s is not running, dumping the pod object we got:", ns, podName)
+			t.Logf("Pod %s/%s is not running, dumping the pod object we got:", ns, podName)
 			b, _ := json.MarshalIndent(p, "", " ")
 			t.Log(string(b))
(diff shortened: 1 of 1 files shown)


pkg/k8s/wait.go
```diff
@@ -250,11 +250,20 @@ func WaitForPodRunningOrFail(ctx context.Context, t feature.T, podName string) {
...
 
+			logs, err := podClient.
+				GetLogs(p.GetName(), &corev1.PodLogOptions{}).
+				Do(ctx).
+				Raw()
+			if err != nil {
+				t.Logf("failed to get pod logs: %w", err)
+			} else {
+				t.Logf("Pod %s/%s logs: %s", string(logs))
+			}
```
cardil, line 265: Could we rewrite this to dump the logs to a report file (within `$ARTIFACTS` dir) and log only a link to the file dumped? Similar to: https://github.com/cardil/knative-reconciler-test/blob/52949cd7a6a198ff682dee561f87829fb227f3b6/pkg/milestone/emitter_log.go#L128-L141

Putting so much information on console makes log browsing a pain.

cardil (changes requested):

pkg/k8s/wait.go
pierDipi: unready pods don't have many logs, anyway I'll revisit this in the future, so closing

/close

PR #288, opened 2022-01-26
Wait for resource deletions in DeleteResources

pierDipi:
This patch makes the method `f.DeleteResources` semantic more
clear by waiting for all resources to be deleted before returning.

This guarantees that once a feature passed to a next phase
resources were actually deleted.

Signed-off-by: Pierangelo Di Pilato <pierdipi@redhat.com>

PR #293, opened 2022-01-27
Auto logging on teardown

pierDipi:
The existing tests framework in eventing has the nice feature
of logging resources that are tracked during a test and that's
helpful for debugging test issues.

This PR is adding support for the same feature, by automatically
prepending logging steps to the teardown phase of every feature.

/hold this PR is based on https://github.com/knative-sandbox/reconciler-test/pull/288

### DIFF
--- a/pkg/environment/logging.go
+++ b/pkg/environment/logging.go
@@ -1,0 +1,34 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package environment
+
+import (
+	"knative.dev/reconciler-test/pkg/feature"
+)
+
+// loggingSteps returns a number of steps that logs ma
+func (mr *MagicEnvironment) loggingSteps() []feature.Step {
+	return []feature.Step{
+		{
+			Name: "Log references",
+			S:    feature.Any,
+			L:    feature.Must,
+			T:    feature.Teardown,
+			Fn:   feature.LogReferences(mr.refs...),
+		},
+	}
+}
--- a/pkg/environment/logging.go
+++ b/pkg/environment/logging.go
@@ -22,13 +22,11 @@ import (
 
 // loggingSteps returns a number of steps that logs ma
 func (mr *MagicEnvironment) loggingSteps() []feature.Step {
-	return []feature.Step{
-		{
-			Name: "Log references",
-			S:    feature.Any,
-			L:    feature.Must,
-			T:    feature.Teardown,
-			Fn:   feature.LogReferences(mr.refs...),
-		},
-	}
+	return []feature.Step{{
+		Name: "Log references",
+		S:    feature.Any,
+		L:    feature.Must,
+		T:    feature.Teardown,
+		Fn:   feature.LogReferences(mr.refs...),
+	}}
 }
--- a/pkg/environment/logging.go
+++ b/pkg/environment/logging.go
@@ -20,7 +20,7 @@ import (
 	"knative.dev/reconciler-test/pkg/feature"
 )
 
-// loggingSteps returns a number of steps that logs ma
+// loggingSteps returns a number of steps that logs environment-managed resources.
 func (mr *MagicEnvironment) loggingSteps() []feature.Step {
 	return []feature.Step{{
 		Name: "Log references",
--- a/pkg/environment/magic.go
+++ b/pkg/environment/magic.go
@@ -294,8 +294,10 @@ func (mr *MagicEnvironment) Test(ctx context.Context, originalT *testing.T, f *f
 		// TODO implement fail fast feature to avoid proceeding with testing if an "expected level" assert fails here
 	}
 
-	// Prepend logging steps to the teardown phase.
-	steps[feature.Teardown] = append(mr.loggingSteps(), steps[feature.Teardown]...)
+	if originalT.Failed() {
+		// Prepend logging steps to the teardown phase.
+		steps[feature.Teardown] = append(mr.loggingSteps(), steps[feature.Teardown]...)
+	}
 
 	for _, s := range steps[feature.Teardown] {
 		s := s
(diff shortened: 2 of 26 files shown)


pkg/feature/logging.go
```diff
@@ -0,0 +1,84 @@
...
+	"knative.dev/pkg/injection/clients/dynamicclient"
+)
+
+func LogReferences(refs ...corev1.ObjectReference) StepFn {
+	return func(ctx context.Context, t T) {
+		for _, ref := range refs {
+			logReference(ref)(ctx, t)
+		}
+	}
+}
```
cardil, line 38: Could we rewrite this method to dump those objects to a report file (within `$ARTIFACTS` dir) and log only a link to the file and number of objects dumped? Similar to: https://github.com/cardil/knative-reconciler-test/blob/52949cd7a6a198ff682dee561f87829fb227f3b6/pkg/milestone/emitter_log.go#L128-L141

Putting everything on console makes log browsing a pain.

pkg/feature/feature.go
```diff
@@ -151,19 +151,54 @@ func (f *Feature) DeleteResources(ctx context.Context, t T) {
...
 		if err != nil && !apierrors.IsNotFound(err) {
-			refFailedDeletion = append(refFailedDeletion, ref)
 			t.Logf("Warning, failed to delete %s/%s of GVR: %+v: %v", ref.Namespace, ref.Name, resource, err)
 		}
 	}
+
+	// refFailedDeletion keeps the failed to delete resources.
+	var refFailedDeletion []corev1.ObjectReference
+
+	err := wait.Poll(time.Second, 4*time.Minute, func() (bool, error) {
```
cardil, line 169: Maybe take timing from `context.Context`?

pkg/environment/logging.go
```diff
@@ -0,0 +1,34 @@
...
+func (mr *MagicEnvironment) loggingSteps() []feature.Step {
+	return []feature.Step{
+		{
+			Name: "Log references",
+			S:    feature.Any,
+			L:    feature.Must,
+			T:    feature.Teardown,
+			Fn:   feature.LogReferences(mr.refs...),
+		},
+	}
```
cardil, line 33:
```suggestion
	return []feature.Step{{
		Name: "Log references",
		S:    feature.Any,
		L:    feature.Must,
		T:    feature.Teardown,
		Fn:   feature.LogReferences(mr.refs...),
	}}
```

pkg/environment/logging.go
```diff
@@ -0,0 +1,34 @@
...
+limitations under the License.
+*/
+
+package environment
+
+import (
+	"knative.dev/reconciler-test/pkg/feature"
+)
+
+// loggingSteps returns a number of steps that logs ma
```
cardil, line 23: The sentence in this comment seams not finished

pkg/environment/magic.go
```diff
@@ -294,6 +294,9 @@ func (mr *MagicEnvironment) Test(ctx context.Context, originalT *testing.T, f *f
 		// TODO implement fail fast feature to avoid proceeding with testing if an "expected level" assert fails here
 	}
 
+	// Prepend logging steps to the teardown phase.
+	steps[feature.Teardown] = append(mr.loggingSteps(), steps[feature.Teardown]...)
```
cardil, line 298: This means that the objects will be dumped even if tests passes. It would be good to dump them only if test failed.

cardil (changes requested):
Nice idea! Looks good apart from few comments. 

Please respond to comments, with your opinion, so we could resolve them.

pkg/environment/magic.go
pierDipi: Logging in tests isn't actually printed in the failed test case if the test doesn't use the `verbose` logging flag

pkg/environment/logging.go
pierDipi: :man_facepalming:

pierDipi, line 23:
```suggestion
// loggingSteps returns a number of steps that logs environemnt managed resources.
```

pkg/feature/feature.go
pierDipi: Unfortunately, there is a cyclic dependency between the packages, `environment` imports `feature` and `feature` can't import `environment` to pull timings, so I wanted to re-org packages in a separate PR

pkg/feature/logging.go
pierDipi: While working at debugging failed tests I found debugging a lot easier and quicker when logs are immediately searchable, navigating artifacts files should be the last resort for really deep insights and weird bugs.

Also, having logs immediately searchable lowers the barrier for new contributors who don't have a clear understanding of where things are, so I'd prefer logging them directly.

pierDipi: This is also the behavior that we already have in eventing core testing libraries, so people are already familiar with that.

pkg/environment/magic.go
pierDipi: However, let me add a check originalT.Failed().

pkg/environment/magic.go
```diff
@@ -294,6 +294,9 @@ func (mr *MagicEnvironment) Test(ctx context.Context, originalT *testing.T, f *f
...
+
```
pierDipi, line 299:
```suggestion
	if originalT.Failed() {
		// Prepend logging steps to the teardown phase.
		steps[feature.Teardown] = append(mr.loggingSteps(), steps[feature.Teardown]...)
	}

```

pkg/environment/magic.go
pierDipi: Done

pkg/feature/logging.go
pierDipi: I disagree as stated in my previous comments, I spent quite a lot of time debugging failed tests, it is slow to look at the uploaded logs.

Even if library users want this behavior (and I don't) they can provide their own implementation of `feature.T` that forwards the logs to a file but this has to be opt-in.

pierDipi: and trust me, every eventing event that is logged in the console has been useful to root cause failed tests, so I don't see them as cluttering the console but just providing enough material to debug tests.

cardil: This been discussed here, and accepted https://github.com/knative/eventing/issues/4592

cardil: I think we should survey the community on how we should proceed with test output.

For example, we are running tests with `-v` to see the non-crucial results, like the unavailability time during upgrades. Dumping too much when it's not needed (test pass) makes it harder to browse the logs, and also make the execution slower.

I'll accept it for now. And we will go back to this in separate PR.  (probably #294)

cardil (approved):
It looks ok to me, apart from https://github.com/knative-sandbox/reconciler-test/pull/293/files#r797742084 (but we can continue on that issue in separate PR)

/lgtm
/approve

cardil (approved):
/lgtm

PR #284, opened 2022-01-17
:bug: Eventshub receiver readiness ensured

cardil:
# Changes

- :bug: Eventshub receiver readiness ensured

/kind bug

Fixes #270
Closes #271

PR #294, opened 2022-01-27
:broom: Logging overhaul

### LINKED ISSUE #295: Logging in reconciler-test is clumsy
The logging in this library is all over the place. The following packages are used to log messages:

- `testing`
- `zap`
- `fmt`
- `log`
- `k8s.io/klog`
 
As this library is test focused, it would be a good idea to feed all logs to `testing` package. The `testing` package loggers automatically works when doing a parallel testing, so the messages will end up in appropriate test report. Also, `testing` loggers print messages only in case of failure, or if `-v` test flag is used.

I think the best solution would be to set up [`zaptest.NewLogger`](https://pkg.go.dev/go.uber.org/zap/zaptest#NewLogger), so that it will feed all messages to appropriate `testing.T` object, and allow using fast, leveled, and structured logging across codebase.

/kind cleanup

cardil:
# Changes

- :gift: Adds the `environment.NewStandardGlobalEnvironment()` to simplify environment creation.
- :broom: Moving init code to `TestMain()` to avoid side effects, while importing a package.
- :broom: Logging overhaul to utilize `zap.Logger` for testing [`zaptest.NewLogger`](https://pkg.go.dev/go.uber.org/zap/zaptest#NewLogger).

/kind cleanup

Fixes #295
Requires #284
Related knative/eventing#4592


cardil:
Looking for feedback for this PR. I think the changes are good, but they contain some breaking changes. If the changes are okay, I could provide additional PRs for downstream projects to merge all of them at once. I'd like to avoid splitting the API just to have a deprecation period, which no one would notice. Better to rewrite dependent projects in one go. At least that's my point of view.

WDYT?

@dsimansk 
@zroubalik 
@dprotaso
@pierDipi 
@matzew
@upodroid
@psschwei
@kvmware
@aliok
@odacremolbap

cardil:
Looking for feedback for this PR. I think the changes are good, but they contain some breaking changes. If the changes are okay, I could provide additional PRs for downstream projects to merge all of them at once. I'd like to avoid splitting the API just to have a deprecation period, which no one would notice. Better to rewrite dependent projects in one go. At least that's my point of view.

WDYT?

/cc @dsimansk 
/cc @zroubalik 
/cc @dprotaso
/cc @pierDipi 
/cc @matzew
/cc @upodroid
/cc @psschwei
/cc @kvmware
/cc @aliok
/cc @odacremolbap

pkg/environment/magic.go
```diff
@@ -106,22 +115,20 @@ func WithPollTimings(interval, timeout time.Duration) EnvOpts {
...
 func Managed(t feature.T) EnvOpts {
-	return func(ctx context.Context, env Environment) (context.Context, error) {
+	cleanup := func(ctx context.Context, env Environment) (context.Context, error) {
 		if e, ok := env.(*MagicEnvironment); ok {
 			e.managedT = t
 		}
 		t.Cleanup(env.Finish)
 		return ctx, nil
 	}
+	return UnionOpts(cleanup, WithTestLogger(t))
```
mgencur, line 128: Nit: the `cleanup` function already does two things:
* registers `t` as `managedT`
* registers Cleanup on `t`
And then we merge it with another one-line funtion WithTestLogger that is not used anywhere else.
I'm wondering if this should be merged into one function that does all of that:
```
func Managed(t feature.T) EnvOpts {
	return func(ctx context.Context, env Environment) (context.Context, error) {
		if e, ok := env.(*MagicEnvironment); ok {
			e.managedT = t
		}
		t.Cleanup(env.Finish)
		return testlog.WithTestLogger(ctx, t), nil
	}
}
```

pkg/environment/images.go
```diff
@@ -17,62 +17,119 @@ limitations under the License.
...
-		}
-		if !exists {
-			packages = append(packages, p)
-		}
+func RegisterPackage(pack ...string) EnvOpts {
+	ri := func(ctx context.Context, env Environment) (context.Context, error) {
+		rk := registeredPackagesKey{}
+		return rk.register(ctx, pack), nil
 	}
+	return UnionOpts(ri, WithImages(map[string]string{}))
```
mgencur, line 46: Why is  `WithImages` called here? It looks unrelated. The function `imageStoreKey.get` returns a new (empty) image store if it's not initialized yet so I think it's not necessary to call it here. It changes the meaning of the whole `RegisterPackage` function. It's not only registering a package but does more. I think it's not necessary to call WithImages here and so the UnionOpts function can be removed as well.

mgencur (commented):
Hey @cardil , overall I like it! And I left a few comments, mostly for cleaning that up a bit.

pkg/logging/context.go
```diff
@@ -20,20 +20,48 @@ import (
...
+			zap.String("test", t.Name()),
+		)),
+	}, opts...)
+	log := zaptest.NewLogger(t, opts...)
+	return logging.WithLogger(ctx, log.Sugar())
 }
 
-func logger() *zap.SugaredLogger {
-	if log, err := zap.NewDevelopment(); err != nil {
+func interimLogger(level zapcore.Level) *zap.SugaredLogger {
```
mgencur, line 61: Nit: Let's call it `defaultLogger` ?

pkg/milestone/emitter.go
```diff
@@ -54,7 +54,7 @@ type Emitter interface {
 	StepFinished(feature string, step *feature.Step, t feature.T)
 	TestSetStarted(featureSet string, t feature.T)
 	TestSetFinished(featureSet string, t feature.T)
-	Finished()
+	Finished(failed bool)
```
pierDipi, line 57: I'd pass a struct for future changes/addition

pkg/environment/images.go
```diff
@@ -17,62 +17,119 @@ limitations under the License.
...
-var packageToImageConfig = map[string]string{}
-var packaged sync.Once
+var (
+	// CurrentImageProducer is the function that will be used to produce the
+	// container images. By default, it is ko.Publish, but can be overridden.
+	CurrentImageProducer = ImageProducer(ko.Publish)
+)
+
+// ImageProducer is a function that will be used to produce the container images.
+type ImageProducer func(ctx context.Context, pack string) (string, error)
```
pierDipi, line 34: are these changes related to logging?
can we break them down into smaller pieces?

cardil: > _are these changes related to logging?_
Yes. The image producer was logging directly to standard output, and was doing that before the test even started.

> _can we break them down into smaller pieces?_
If we break this up to smaller pieces, it will be even harder to merge all of this at once. Those are breaking changes, so my proposal for now is to prepare a set of PRs (this, plus downstream PRs) and merge all at once.

mgencur:
> But, what do you think about my proposal to merge it, by providing rewrite PRs for downstream projects? This PR break downstream projects compilation. Do you think that strategy is acceptable?
I think it makes sense to create those PRs in advance. And when the current PR is merged you'd run ./hack/updated-deps.sh manually and update the downstream PRs. We could then merge them with tests passing.

test/e2e/eventshub/receiver_readiness.go
```diff
@@ -50,6 +50,10 @@ const ubi8Image = "registry.access.redhat.com/ubi8/ubi"
 // ReceiverReadiness checks the readiness of the Events Hub receiver.
 func ReceiverReadiness() *feature.Feature {
 	f := feature.NewFeature()
+
+	// Ensure the eventshub image is registered.
+	f.ConfigureEnvironment(eventshub.Images())
```
pierDipi, line 55: Do we expect every test to call this method when eventshub is used?

I don't get the relation with logging here.

pkg/environment/standard.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	for _, opt := range opts {
+		cfg = opt(cfg)
+	}
+	return cfg
+}
+
+type commandlineFlags struct{}
+
+func (c commandlineFlags) Set(ctx context.Context) *flag.FlagSet {
```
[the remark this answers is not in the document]
cardil: Good point. The `Get` would be a better name.

The overall idea for this change is to allow passing `*flag.FlagSet` differently. Still, the default uses the [`flag.CommandLine`](https://pkg.go.dev/flag#pkg-variables).

test/e2e/eventshub/main_test.go
```diff
@@ -39,24 +36,9 @@ var global environment.GlobalEnvironment
...
-	// features to pull Kubernetes clients or the test environment out of the
-	// context passed in the features.
-	ctx, startInformers := injection.EnableInjectionOrDie(
-		logging.WithTestLogger(nil), nil) // nolint
-	startInformers()
-
-	// global is used to make instances of Environments, NewGlobalEnvironment
+	// global is used to make instances of Environments, NewStandardGlobalEnvironment
 	// is passing and saving the client injection enabled context for use later.
-	global = environment.NewGlobalEnvironment(ctx)
```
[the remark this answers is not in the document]
cardil: Why? To what exactly are you referring?

I just wrap the initialization in a function. It still does the same job.

test/e2e/eventshub/receiver_readiness.go
mgencur: I guess it would be good to move the image registration directly into `eventshub.Install()`.

pkg/environment/images.go
cardil: I need the `WithImages` and `RegisterPackage` to be called early so the structs they manage (although empty) are within the associated `context.Context` object - https://github.com/cardil/knative-reconciler-test/blob/bdc1b811fcfc957b1071d0b38e40a3c01e9abe8e/pkg/environment/magic.go#L145.

This allows the tests to register packages, and resolve images at later stages by using the new `(f *Feature) ConfigureEnvironment(opt ...Option)` call. For example: https://github.com/cardil/knative-reconciler-test/blob/bdc1b811fcfc957b1071d0b38e40a3c01e9abe8e/test/example/config/echo/echo.go#L37-L43

test/e2e/eventshub/receiver_readiness.go
cardil: Not test, but feature.

The relation with logging is non-direct. The manifest parsing functions has been rewritten to utilize the `logging.FromContext(context.Context)` calls. That also pulled the need to rewrite the image stuff (to utilize context as well, plus a proper logger ofc).

I thought of providing a global variable/function, say "auto-register-packages", and an initializer that would be enabled by default. That would allow keeping current behavior of automagically registering images just by using import. I don't include that code in this PR because I believe such side effects are bad code in my opinion.

Quote from [`github.com/leighmcculloch/gochecknoinits`](https://github.com/leighmcculloch/gochecknoinits):

> _Init functions cause an import to have a side effects, and side effects are hard to test, reduce readability and increase the complexity of code._

If you strongly feel, we should include that auto-registration feature, I could amend this PR with appropriate code. But, I'm reluctant to doing that.

pkg/environment/images.go
mgencur: I still don't get why the `imageStoreKey` needs to be initialized early with an empty map. I don't see any relation between the initialization and later usage in `ConfigureEnvironment(opt ...Option)`. It's not used used there. And when someone wants to use a specific image set they can call `WithImages` by themselves later. It just needs to be done before installing the feature here: https://github.com/knative-sandbox/reconciler-test/pull/294/files#diff-3aff8e97ed3181bb28543c9b61577c312f24086c81b2b71a0b64eaf5865e77d5R49

test/e2e/eventshub/receiver_readiness.go
mgencur: Why isn't `eventshub.Images()` encapsulated in `eventshub.Install()`? Perhaps I'm missing something.

cardil: I have tried to encapsulate the `eventshub.Images()` within `eventshub.Install()`, but that turns out to be problematic.

The `eventshub.Install()` uses `feature.StepFn` which expects the user to provide a configured `context.Context`. That configured `context.Context` will not propagate to adjacent `.Test(Feature)`.

That's why I've added new `Feature` API of [`.ConfigureEnvironment(opt ...Option)`](https://github.com/cardil/knative-reconciler-test/blob/bdc1b811fcfc957b1071d0b38e40a3c01e9abe8e/pkg/feature/feature.go#L216-L218). Using it, the context is being reconfigured before it is passed to `feature.StepFn`.

Saying the above, I need to reevaluate those statements. I have added the following invocation: https://github.com/cardil/knative-reconciler-test/blob/bdc1b811fcfc957b1071d0b38e40a3c01e9abe8e/pkg/environment/magic.go#L145 and that should configure context with mutable structs. Those mutable structs should be present on every Test/Feature, so theoretically it may be possible to simplify this and move the `eventshub.Images()` to the `eventshub.Install()`. Let me check this...

mgencur: Thanks for double-checking this. It would be nice to encapsulate in that function.

test/e2e/eventshub/main_test.go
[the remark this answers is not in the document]
cardil: Fixed in https://github.com/knative-sandbox/reconciler-test/pull/294/commits/16e76f1c783185734268c2f1e44a7f1fbd4201a6

pkg/environment/magic.go
```diff
@@ -34,27 +34,28 @@ import (
 	"knative.dev/reconciler-test/pkg/state"
 )
 
-func NewGlobalEnvironment(ctx context.Context) GlobalEnvironment {
-	fmt.Printf("level %s, state %s, feature %s\n\n", l, s, *f)
-
+func NewGlobalEnvironment(ctx context.Context, starters ...func()) GlobalEnvironment {
```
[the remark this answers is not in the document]
cardil: Fixed in 86f8703a05f9e7b4b4f0ecf2dc3fa92a8e887854

pkg/environment/standard.go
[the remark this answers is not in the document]
cardil: Fixed in 4d7ba10bd88707dfa7269deb52144a30e94c7d5b

pkg/logging/context.go
cardil: Fixed in c034e47e6671395fd2a6b05c1bb88464bbf1c82f

pkg/milestone/emitter.go
cardil: Fixed in 82e496323312aab5b5d9414135ba1efaa23eb66d. Hopefully...

test/e2e/eventshub/receiver_readiness.go
cardil: After some little changes, I've managed to move the image registration to within the installation method.

So, the Feature API of [.ConfigureEnvironment(opt ...Option)](https://github.com/cardil/knative-reconciler-test/blob/bdc1b811fcfc957b1071d0b38e40a3c01e9abe8e/pkg/feature/feature.go#L216-L218) will not be needed.

Fixed in 28d2f63

pkg/environment/magic.go
```diff
@@ -130,32 +145,44 @@ func (mr *MagicGlobalEnvironment) Environment(opts ...EnvOpts) (context.Context,
 		s:            mr.FeatureState,
 		featureMatch: mr.FeatureMatch,
 
-		images:    images,
 		namespace: namespace,
 	}
 
 	ctx := ContextWith(mr.c, env)
 
+	opts = append(opts, RegisterPackage( /* non, to initialize */ ))
```
cardil, line 153: While doing https://github.com/knative-sandbox/reconciler-test/pull/294#discussion_r920267067 I've noticed this registration, isn't in the best place.

By calling it here, each Golang test would have a clean context. This means the images will be built for every test, which isn't ideal - that's also a direct regression in behavior for this PR.

I think by moving this registration to global environment initialization, we could have the same behavior as before - one package, one build.

/hold

cardil: After further looking into this (see 289e872c6fbf37c69b63ca39a1e5eba5f72d7c84), I came to a conclusion the problem here is larger. I'll try to capture it below:

1. I wrote the image store (holds built images references), and packages store (holds registered packages to be built) as mutable objects. That was intentional, to prevent excessive, non-necessary image builds.
2. When those stores are contained within the `context.Context` object that is shared by different tests, and those tests are parallelized, race detection fires.
3. Even when this issue is somehow solved, there is a problem with the `WithImages` function. The `WithImages` function is supposed to overwrite the images, and theirs references. If it's used in test A, currently it will influence the test B, this shouldn't happen.

All in all, I think the solution to above could be:

- Use some producer-consumer channel-backed solution for `RegisterPackages` and `ProduceImages` to satisfy the race detection checker
- Change the `WithImages` function to return a new `context.Context` with a new image store assigned to it, instead of replacing contents of an already assigned image store. This will create test specific `ctx` object that will not influence other tests.

cardil: Race problems fixed with 32bfa2db7d91d76019490b9fd19d5d9959f098d9

cardil: The `WithImages` problem fixed with 169264f8ba2395f1e860194c2c32330f18b19e77

/unhold

mgencur:
Hi @cardil , the PR looks good now! It's already approved. Do you want to create the dependent PRs before we put lgtm on this? Also, @pierDipi any other comments?

cardil:
> Hi @cardil , the PR looks good now! It's already approved. Do you want to create the dependent PRs before we put lgtm on this? Also, @pierDipi any other comments?

Great! Thanks a lot for the review @mgencur!

Yes. I would like to create now those downstream PRs before merging this. Than we could merge all at once to avoid breakages.

pierDipi:
> > Hi @cardil , the PR looks good now! It's already approved. Do you want to create the dependent PRs before we put lgtm on this? Also, @pierDipi any other comments?
> 
> Great! Thanks a lot for the review @mgencur!
> 
> Yes. I would like to create now those downstream PRs before merging this. Than we could merge all at once to avoid breakages.

Thanks! this will also help me understand the API changes better.

mgencur:
Alright, let's put hold on this until those PRs are created to prevent accidental merging.
/hold

cardil:
Downstream PRs tracking - [dependents list](https://deps.dev/go/knative.dev%2Freconciler-test/v0.0.0-20220412165608-994f0c3fab62/dependents):

| Package       | PR            | State         |
| ------------- | ------------- | ------------- |
| [`knative.dev/control-protocol`](https://github.com/knative-sandbox/control-protocol) | https://github.com/knative-sandbox/control-protocol/pull/197  | :question: wip |
| [`knative.dev/eventing`](https://github.com/knative/eventing) | https://github.com/knative/eventing/pull/6451  | :heavy_check_mark: tests passing |
| [`knative.dev/eventing-autoscaler-keda`](https://github.com/knative-sandbox/eventing-autoscaler-keda) | missing | :red_circle: missing |
| [`knative.dev/eventing-kafka`](https://github.com/knative-sandbox/eventing-kafka) | https://github.com/knative-sandbox/eventing-kafka/pull/1213 | :heavy_check_mark: tests passing |
| [`knative.dev/eventing-kafka-broker`](https://github.com/knative-sandbox/eventing-kafka-broker) | https://github.com/knative-sandbox/eventing-kafka-broker/pull/2413 | :question: wip |
| [`knative.dev/eventing-natss`](https://github.com/knative-sandbox/eventing-natss) | missing | :red_circle: missing |
| [`knative.dev/eventing-rabbitmq`](https://github.com/knative-sandbox/eventing-rabbitmq) | missing | :red_circle: missing |
| [`knative.dev/kn-plugin-event`](https://github.com/knative-sandbox/kn-plugin-event) | https://github.com/knative-sandbox/kn-plugin-event/pull/203 | :heavy_check_mark: tests passing |

cardil:
We need someone with review/approvers powers to start merging process:

@knative-sandbox/technical-oversight-committee 
@knative-sandbox/productivity-writers 
@knative-sandbox/eventing-writers 
@knative-sandbox/knative-release-leads

dsimansk:
@cardil happy to help! There's a hold label so feel to lift it as needed.

/approve
/lgtm

cardil:
Staring the merge rollout

/unhold

PR #300, opened 2022-02-09
Features are stored by reference in a FeatureSet

### LINKED ISSUE #299: Feature's DeleteResources doesn't work
When a Feature is being planned and `f.DeleteResources` is called (like [here](https://github.com/knative-sandbox/reconciler-test/blob/main/test/example/echo_feature.go#L78) and [here](https://github.com/knative/eventing/blob/main/test/rekt/features/broker/control_plane.go#L392)), the feature being referenced doesn't have any Object References.

When feature sets are [run](https://github.com/knative-sandbox/reconciler-test/blob/main/pkg/environment/magic.go#L321-L325), it will make a copy of the Feature value and run the test. The feature is stored in the context as seen [here](https://github.com/knative-sandbox/reconciler-test/blob/main/pkg/environment/magic.go#L242) and gets retrieved in other areas where ObjectReferences are attached to the feature. 

When `DeleteResources` is ultimately called, the feature being referenced is the old incomplete one while the completed one exists in the context. A simple solution would be to retrieve the Feature from the context in `DeleteResources` and switch the method to a function to avoid confusion. But since this change will have downstream impact, I'll defer to the maintainers.

/kind bug

gab-satchi:
# Changes
- Features are stored by reference in a FeatureSet

- 🐛  fix issue with cleaning up resources where the original Feature is lost as it's passed by value. 

/kind bug

Fixes #299 

/assign @pierDipi  Since you worked on deleteResources recently :)

### DIFF
--- a/pkg/feature/feature.go
+++ b/pkg/feature/feature.go
@@ -93,7 +93,7 @@ func NewFeature() *Feature {
 // FeatureSet is a list of features and feature set name.
 type FeatureSet struct {
 	Name     string
-	Features []Feature
+	Features []*Feature
 }
 
 // StepFn is the function signature for steps.
@@ -139,8 +139,7 @@ func (f *Feature) References() []corev1.ObjectReference {
 // Use References to get the undeleted resources.
 //
 // Expected to be used as a StepFn.
-func DeleteResources(ctx context.Context, t T) {
-	f := FromContext(ctx)
+func (f *Feature) DeleteResources(ctx context.Context, t T) {
 	dc := dynamicclient.Get(ctx)
 	for _, ref := range f.References() {
 
@@ -205,7 +204,7 @@ func DeleteResources(ctx context.Context, t T) {
 
 var (
 	// Expected to be used as a StepFn.
-	_ StepFn = DeleteResources
+	_ StepFn = (&Feature{}).DeleteResources
 )
 
 // Setup adds a step function to the feature set at the Setup timing phase.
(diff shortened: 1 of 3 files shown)


gab-satchi:
These changes will break downstream tests. I'm not sure of the process used in a case like this

gabo1208 (approved):

pkg/feature/feature.go
```diff
@@ -139,7 +139,8 @@ func (f *Feature) References() []corev1.ObjectReference {
 // Use References to get the undeleted resources.
 //
 // Expected to be used as a StepFn.
-func (f *Feature) DeleteResources(ctx context.Context, t T) {
+func DeleteResources(ctx context.Context, t T) {
```
pierDipi, line 142: I'm confused on why `f != FromContext(ctx)` here?

gab-satchi: Yeah I'm good to make that switch. It would have (potentially) have larger changes needed downstream but it would be the correct thing to do. Will make the changes, Thanks!

gab-satchi:
hey @pierDipi updated featureset to track features by reference

pierDipi (approved):
/lgtm

I'm not an approver here, so we need someone to approve this

matzew:
@gab-satchi looks like this PR has some issues, and breaks updates. See PRs like: https://github.com/knative/eventing/pull/6173

gab-satchi:
@matzew thanks yeah those updates were expected when this change was made since they all use the featureset.

PR #302, opened 2022-02-14
Add EventHub readiness probe only in the receiver role

### LINKED ISSUE #301: Readiness probe on sender causing failures
Readiness probes were introduced in #284. While that fixed the issue for receivers, senders have become more flakey as a result. 

The sender and receiver share the same [Install](https://github.com/knative-sandbox/reconciler-test/blob/main/pkg/eventshub/resources.go#L49) function. And it's now including a Readiness probe for senders as well as receivers. Senders don't respond to that endpoint at all so the pod never becomes Ready. It will still pass the test's step if the pod happens to complete sending events before the timeout as we wait for Ready or Done states. But if the Sender takes longer than the timeout to send all the events, it will fail the step and the rest of the test. 

Fix would be to selectively include Readiness probe only on the Receivers.

pierDipi:
Senders don't start any server to receive readiness probes.

# Changes

- Add EventHub readiness probe only in the receiver role

/kind bug

Fixes #301 

**Release Note**

```release-note
Add EventHub readiness probe only in the receiver role
```

### DIFF
--- a/pkg/eventshub/103-pod.yaml
+++ b/pkg/eventshub/103-pod.yaml
@@ -26,10 +26,12 @@ spec:
     - name: eventshub
       image: {{ .image }}
       imagePullPolicy: "IfNotPresent"
+      {{ if .withReadiness }}
       readinessProbe:
         httpGet:
           port: 8080
           path: /health/ready
+      {{ end }}
       env:
         - name: "SYSTEM_NAMESPACE"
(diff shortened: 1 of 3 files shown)


gab-satchi (commented):
thanks @pierDipi 

/lgtm

cardil (approved):
/lgtm

PR #360, opened 2022-07-06
Dump collected events on failed matches

pierDipi:
Debugging failed matches is complicated as it is without getting
all events that were collected.


matzew (approved):
/lgtm
/approve

PR #371, opened 2022-08-08
Fix String() call on nil events on event sender

gabo1208:
# Changes

- 🐛  with the new ce checks on the broker conformance test suite sometimes… there are nil events here that are trying to get converted to string, this is a nil check to avoid the event source to error during the test runs

/kind bug

### DIFF
--- a/pkg/eventshub/sender/sender.go
+++ b/pkg/eventshub/sender/sender.go
@@ -179,9 +179,13 @@ func Start(ctx context.Context, logs *eventshub.EventLogs, clientOpts ...Option)
 			return err
 		}
 
+		eventString := ""
+		if event != nil {
+			eventString = event.String()
+		}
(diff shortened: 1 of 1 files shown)


pkg/eventshub/sender/sender.go
```diff
@@ -179,9 +179,13 @@ func Start(ctx context.Context, logs *eventshub.EventLogs, clientOpts ...Option)
 			return err
 		}
 
+		eventString := ""
```
pierDipi, line 182: nit, so that it's clear that the request wasn't an event
```suggestion
		eventString := "unknown"
```

matzew: I like this suggestion, @gabo1208

pierDipi (approved):
/lgtm
/approve

PR #394, opened 2022-09-08
Run steps with the same timing in parallel as documented

pierDipi:
The `README` is explicity documenting that timings
(Setup, Requirement, etc) runs in order but steps with the
same timing runs in parallel while currently, this is not the
case and it is slowing test execution time down.

# Changes

- Run steps with the same timing in parallel

**Release Note**

```release-note
Run steps with the same timing in parallel as documented.
```

### DIFF
--- a/pkg/environment/execution.go
+++ b/pkg/environment/execution.go
@@ -2,21 +2,24 @@ package environment
 
 import (
 	"context"
-	"sync"
 	"testing"
 
 	"knative.dev/reconciler-test/pkg/feature"
 )
 
-func categorizeSteps(steps []feature.Step) map[feature.Timing][]feature.Step {
-	res := make(map[feature.Timing][]feature.Step, 4)
+type CategorizedSteps struct {
+	Timing feature.Timing `json:"timing"`
+	Steps  []feature.Step `json:"steps"`
+}
 
-	res[feature.Setup] = filterStepTimings(steps, feature.Setup)
-	res[feature.Requirement] = filterStepTimings(steps, feature.Requirement)
-	res[feature.Assert] = filterStepTimings(steps, feature.Assert)
-	res[feature.Teardown] = filterStepTimings(steps, feature.Teardown)
+func categorizeSteps(steps []feature.Step) (map[feature.Timing][]feature.Step, []CategorizedSteps) {
+	stepsByTiming := make(map[feature.Timing][]feature.Step, 4)
 
-	return res
+	for _, timing := range feature.Timings() {
+		stepsByTiming[timing] = filterStepTimings(steps, timing)
+	}
+
+	return stepsByTiming, stepsSorter(stepsByTiming).byTiming()
 }
 
 func filterStepTimings(steps []feature.Step, timing feature.Timing) []feature.Step {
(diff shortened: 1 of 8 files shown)


mgencur:
> but steps with the
same timing runs in parallel while currently

Can you point me to the docs which say the above? I only found docs that speak about running individual steps in order.
I think your change is problematic at least for the Setup phase. Some tests will definitely count on running individual Setup steps in order (i.e. all Setup steps one by one, not in parallel). For example, I can see [this in eventing-kafka-broker](https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/test/e2e_new/broker_delete_config_map_test.go#L65-L70): 
```
        f.Setup("install broker", broker.Install(
		brokerName,
		broker.WithBrokerClass(broker.EnvCfg.BrokerClass),
	))
	f.Setup("broker is ready", broker.IsReady(brokerName))
	f.Setup("broker is addressable", broker.IsAddressable(brokerName))
```
With your parallel execution the "broker is ready" could actually run before "install broker" (in theory), and the test would fail.
I think running things in parallel within certain phase/timing is possible only for "Requirement" and "Assert". Not for `Setup` and `Teardown` where people might expect serial execution and depend on the order.

pkg/environment/execution.go
```diff
@@ -2,21 +2,24 @@ package environment
...
+type CategorizedSteps struct {
+	Timing feature.Timing `json:"timing"`
+	Steps  []feature.Step `json:"steps"`
+}
 
-	res[feature.Setup] = filterStepTimings(steps, feature.Setup)
-	res[feature.Requirement] = filterStepTimings(steps, feature.Requirement)
-	res[feature.Assert] = filterStepTimings(steps, feature.Assert)
-	res[feature.Teardown] = filterStepTimings(steps, feature.Teardown)
+func categorizeSteps(steps []feature.Step) (map[feature.Timing][]feature.Step, []CategorizedSteps) {
```
mgencur, line 15: This function creates redundancy as it basically returns just a list of steps the are also returned as the first parameter as map. I think it might be clearer to just return the map and then use it directly in the loop in magic.go. Then the whole stepSorter would be unnecessary:
```
for timing := range stepsMap {
			steps := stepsMap[timing]
			// Prepend logging steps to the teardown phase when a previous timing failed.
			if skipOnError && timing == feature.Teardown {
				steps = append(mr.loggingSteps(), steps...)
			}

			// Teardown are executed always, no matter their level and state
			if timing == feature.Teardown {
				skipOnError = false
			}

			originalT.Logf("Running %d steps for timing %s\n\n", len(steps), timing.String())

			t.Run(timing.String(), func(t *testing.T) {
				// no parallel, various timing steps run in order: setup, requirement, assert, teardown

				if skipOnError {
					t.Skipf("Skipping steps for timing %s due to failed previous timing\n", timing.String())
					return
				}

				for _, s := range steps {
					s := s
					if mr.shouldFail(&s) {
						mr.execute(ctx, t, f, &s)
					} else {
						mr.executeOptional(ctx, t, f, &s)
					}
				}
			})

			if t.Failed() {
				// skip the following timing since curring timing failed
				skipOnError = true
			}
		}
```

pkg/environment/magic.go
```diff
@@ -295,74 +296,51 @@ func (mr *MagicEnvironment) Test(ctx context.Context, originalT *testing.T, f *f
...
-			break                   // No need to continue the setup
-		}
-	}
+	originalT.Run(f.Name, func(t *testing.T) {
 
-	for _, s := range steps[feature.Requirement] {
-		s := s
+		for _, steps := range sortedSteps {
+			// Prepend logging steps to the teardown phase when a previous timing failed.
+			if skipOnError && steps.Timing == feature.Teardown {
```
mgencur, line 309: What about moving this a few lines below where you already check the phase being Teardown, for clarity:
```
			if timing == feature.Teardown {
				if skipOnError {
					// Prepend logging steps to the teardown phase when a previous timing failed.
					steps = append(mr.loggingSteps(), steps...)
				}
                                // Teardown are executed always, no matter their level and state
				skipOnError = false
			}
```

mgencur: Minor point, maybe the `skipOnError` variable could be named `skipNext` or just `skip`. The OnError suggest this is a general flag that is set before the whole test suite and dictates if steps should be skipped on error. But here we actually change the flag in each iteration. It's kind of confusing to me and I had to read carefully to understand its meaning.

mgencur (commented):
Hi @pierDipi , I think it would be benefial to run steps within Assert and Requirements in parallel but not for Setup and Teardown. Please see my comments. Please let me know if it makes sense.

pierDipi:
> > but steps with the
> > same timing runs in parallel while currently
> 
> Can you point me to the docs which say the above? I only found docs that speak about running individual steps in order. I think your change is problematic at least for the Setup phase. Some tests will definitely count on running individual Setup steps in order (i.e. all Setup steps one by one, not in parallel). For example, I can see [this in eventing-kafka-broker](https://github.com/knative-sandbox/eventing-kafka-broker/blob/main/test/e2e_new/broker_delete_config_map_test.go#L65-L70):
> 
> ```
>         f.Setup("install broker", broker.Install(
> 		brokerName,
> 		broker.WithBrokerClass(broker.EnvCfg.BrokerClass),
> 	))
> 	f.Setup("broker is ready", broker.IsReady(brokerName))
> 	f.Setup("broker is addressable", broker.IsAddressable(brokerName))
> ```
> 
> With your parallel execution the "broker is ready" could actually run (or at least finish executing) before "install broker" (in theory), and the test would fail. I think running things in parallel within certain phase/timing is possible only for "Requirement" and "Assert". Not for `Setup` and `Teardown` where people might expect serial execution and depend on the order.

here: https://github.com/knative-sandbox/reconciler-test#features

> Features have several phases that can be registered, each phase is executed in order, but there are no order guarantees for steps in the same phase. If strict ordering is required it is recommended to break that into an independent feature that is tested on an environment in order required.

> With your parallel execution the "broker is ready" could actually run (or at least finish executing) before "install broker"

Not really, broker is ready as it is implemented it waits the broker to be ready

mgencur:
@pierDipi OK. I can see it now: "but there are no order guarantees for steps in the same phase".

pkg/environment/execution.go
pierDipi: Maps are not ordered while looping over them, so we might end up running teadown before setup

pkg/environment/execution.go
```diff
@@ -29,56 +32,52 @@ func filterStepTimings(steps []feature.Step, timing feature.Timing) []feature.St
...
 
 		// Perform step.
-		s.Fn(internalCtx, internalT)
+		s.Fn(internalCtx, ft)
 	})
+}
 
-	// Wait for the test to execute before spawning the next one
-	wg.Wait()
+type stepsSorter map[feature.Timing][]feature.Step
```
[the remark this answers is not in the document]
mgencur: The sorting is actually done earlier inside categorizeSteps:
```
        for _, timing := range feature.Timings() {
		stepsByTiming[timing] = filterStepTimings(steps, timing)
	}
```
So, maybe just move this part into `byTiming()` and then the names would be alright:

pkg/environment/execution.go
mgencur: In that case, the loop could just start as this:
```
		for _, timing := range feature.Timings() {
			steps := stepsMap[timing]
```
...to save some code.

pkg/environment/execution.go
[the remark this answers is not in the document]
pierDipi: I didn't get what you mean with "So, maybe just move this part into byTiming()", which part?

sort here is referring to "sort the map into a ordered (and traverse-able with range) list"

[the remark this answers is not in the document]
mgencur: I mean the code on [these lines](https://github.com/knative-sandbox/reconciler-test/pull/394/files#diff-e7ea6dc5be093ec9709c64c1e909631112f2e04187cd610f1aaaf1098afbbad2R18-R20) actually does the sorting, so it should be part of byTiming. But please see my other comment. Maybe it's not necessary if you just loop this way https://github.com/knative-sandbox/reconciler-test/pull/394/files#r1071159257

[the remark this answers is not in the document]
pierDipi: I see now, thanks, good idea

mgencur:
I like it now. Thanks @pierDipi 

/lgtm

evankanderson:
Hi,

This seems to be breaking tests in https://github.com/knative-sandbox/control-protocol (at least), and managed to just sneak in under the wire for the `pkg` freeze (which includes `reconciler-test` and `control-protocol`).

Example failure: https://github.com/knative-sandbox/control-protocol/actions/runs/3944375211/jobs/6750204201

Any help debugging would be appreciated.

pierDipi:
I'm trying a fix https://github.com/knative-sandbox/control-protocol/pull/246 and I saw you sent a similar one

pierDipi:
worst case, we can revert and merge this after the release so that we have more time

PR #432, opened 2023-01-17
Fix prober concurrent access

pierDipi:
Multiple go routines can access the prober concurrently,
therefore we need to guard its field access with mutexes.

### DIFF
--- a/test/example/prober_feature.go
+++ b/test/example/prober_feature.go
@@ -59,14 +59,11 @@ func ProberFeatureWithDrop() *feature.Feature {
 	prober := eventshub.NewProber()
 	prober.ReceiversRejectFirstN(5)
 	prober.ReceiversRejectResponseCode(429)
-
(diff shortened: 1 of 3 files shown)


test/example/prober_feature.go
```diff
@@ -59,14 +59,11 @@ func ProberFeatureWithDrop() *feature.Feature {
 	prober := eventshub.NewProber()
 	prober.ReceiversRejectFirstN(5)
 	prober.ReceiversRejectResponseCode(429)
-
-	// Configured the sender for how many events it will be sending.
```
matzew, line 63: why removing the comment?

pierDipi: It is not adding that much value to the method name/call tbh

matzew: It's also not hurting, especially on an example

PR #434, opened 2023-01-18
Adds ability for resources to override namespace in template config.

gab-satchi:
- 🎁  Resources can be created in namespaces other than the test namespace


pierDipi (approved):
/lgtm
/approve

PR #271, opened 2022-01-05
:bug: Reproducer: EventsHub receiver insufficient readiness

cardil:
# Changes

- :bug: A reproducer test for EventsHub receiver insufficient readiness #270 

/kind bug

Reproduces #270

PR #438, opened 2023-01-20
Drop wait for service ready that creates a job for each eventshub

pierDipi:
Each receiver eventshub was creating a job to verify service readiness, this is clearly reinventing the weel since k8s has already means to populate service endpoints only when the pods are ready and running, so `WaitForServiceEndpointsOrFail` is already enough to very pod readiness.


mgencur:
I think this was an old workaround for some behaviour of K8S. @cardil will know better.

pierDipi:
Yes, but from the changes associated with the PR for that issue we only need:
```
      readinessProbe:
        httpGet:
          port: 8080
          path: /health/ready
```

cardil:
Re @pierDipi:

> Yes, but from the changes associated with the PR for that issue we only need

Unfortunately, that's not true. The readiness of the pod is different from the readiness of the service. The service readiness depends on iptables propagation on given node. I often got an outcome in which the pod was ready, but service wasn't which ends up with error: `Connection refused` or `Trying 10.223.247.217...`. That was the case when you have an external sender, like we do have in kn-plugin-event case.

/hold

cardil (changes requested):
/hold

pierDipi:
@cardil can you provide evidence in what you're saying?

```
The kubelet uses readiness probes to know when a container is ready to start accepting traffic. A Pod is considered ready when all of its containers are ready. One use of this signal is to control which Pods are used as backends for Services. When a Pod is not ready, it is removed from Service load balancers.
```

pierDipi:
https://github.com/kubernetes/kubernetes/blob/b0ed87078e4042c12e456c8427dee0fb947a724c/pkg/proxy/endpoints.go#L59-L68

```
	// Serving indiciates whether this endpoint is ready regardless of its terminating state.
	// For pods this is true if it has a ready status regardless of its deletion timestamp.
	// This is only set when watching EndpointSlices. If using Endpoints, this is always
	// true since only ready endpoints are read from Endpoints.
	Serving bool
	// Terminating indicates whether this endpoint is terminating.
	// For pods this is true if it has a non-nil deletion timestamp.
	// This is only set when watching EndpointSlices. If using Endpoints, this is always
	// false since terminating endpoints are always excluded from Endpoints.
	Terminating bool
```

pierDipi:
Stackoverflow is not always authoritative answer and it is also referring to rolling updates (which we don't need and also there are many ways to handle them properly without dropping requests)

pierDipi:
It is also mentioned here what's the actual behavior:
https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/580-pod-readiness-gates

> Pod readiness indicates whether the pod is ready to serve traffic. Pod readiness is dictated by kubelet with user specified readiness probe. On the other hand, pod readiness determines whether pod address shows up on the address list on related endpoints object.

and we're clearly not in this case at all for e2e tests:

> For example, during deployment rolling update, a new pod becomes ready. On the other hand, service, network policy and load-balancer are not yet ready for the new pod due to whatever reason (e.g. slowness in api machinery, endpoints controller, kube-proxy, iptables or infrastructure programming). This may cause service disruption or lost of backend capacity. In extreme cases, if rolling update completes before any new replacement pod actually start serving traffic, this will cause service outage.

pierDipi:
Tests results are not available anymore

cardil:
> Tests results are not available anymore

Well, I didn't introduce that code without a reason. As you can see, the `WaitForServiceReadyOrFail` invocation is added after the `WaitForServiceEndpointsOrFail`. After `WaitForServiceEndpointsOrFail` the K8s should guarantee the service is operational, but unfortunately, it won't. Without `WaitForServiceReadyOrFail`, from time to time, I had failures to connect to the service from a sender pod running on the same namespace. That was either: `Connection refused` or `Trying 10.223.247.217...` type of message.

See also: https://github.com/knative-sandbox/reconciler-test/pull/271#issuecomment-1012129577

cardil:
I think running the `TestEventsHubReceiverReadiness` test in loop should, eventually, reproduce this fault.

test/e2e/eventshub/receiver_readiness_test.go
```diff
@@ -1,49 +0,0 @@
...
-import (
-	"testing"
-
-	"knative.dev/reconciler-test/pkg/environment"
-	"knative.dev/reconciler-test/pkg/k8s"
-	"knative.dev/reconciler-test/pkg/knative"
-	"knative.dev/reconciler-test/test/e2e/eventshub"
-)
-
-func TestEventsHubReceiverReadiness(t *testing.T) {
```
cardil, line 31: I don't see a reason why you are removing this test. It was added just to make sure the eventshub receiver readiness is properly handled. It should be able to reproduce the failure.

cardil (changes requested):

cardil:
As you can see: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative-sandbox_reconciler-test/438/integration-tests_reconciler-test_main/1616497432672079872 recreated that failure nicely!

pierDipi:
/close

Still problematic, so let's keep

PR #441, opened 2023-01-20
Handle already exists error in eventshub's RBAC install

pierDipi:
we might get already exists when multiple eventshub
are installed in parallel.


gab-satchi (approved):
/lgtm

PR #445, opened 2023-01-20
Protect references collection with mutex for concurrent access

pierDipi:
As per title.


gab-satchi (approved):
/lgtm

PR #447, opened 2023-01-23
Fix prober receiver install options

pierDipi:
When calling ReceiverInstall options parameter were appended to the the global prober receiver options which causes misconfiguration, evident when using the prober to install different receivers with different configurations.

### DIFF
--- a/pkg/eventshub/prober.go
+++ b/pkg/eventshub/prober.go
@@ -131,10 +131,10 @@ func (p *EventProber) ReceiversHaveResponseDelay(delay time.Duration) {
 func (p *EventProber) ReceiverInstall(prefix string, opts ...EventsHubOption) feature.StepFn {
 	name := feature.MakeRandomK8sName(prefix)
 	p.setShortNameToName(prefix, name)
-	p.appendReceiverOptions(opts...)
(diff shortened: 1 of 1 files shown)


pkg/eventshub/prober.go
```diff
@@ -131,10 +131,10 @@ func (p *EventProber) ReceiversHaveResponseDelay(delay time.Duration) {
 func (p *EventProber) ReceiverInstall(prefix string, opts ...EventsHubOption) feature.StepFn {
 	name := feature.MakeRandomK8sName(prefix)
 	p.setShortNameToName(prefix, name)
-	p.appendReceiverOptions(opts...)
-	p.appendReceiverOptions(StartReceiver)
+	opts = append(opts, p.getReceiverOptions()...)
```
gab-satchi, line 134: Should the global options be the first in the slice? So individual receivers can override the global config?

pierDipi: Yes, that's better

pierDipi: Done

gab-satchi (approved):
/lgtm

PR #448, opened 2023-01-23
Keep polling when endpoint is not present

pierDipi:
Endpoints are created asynchronously from the Service creation, so we need to keep polling when endpoints are not present.


gab-satchi (approved):
/lgtm

PR #458, opened 2023-02-09
Add Pod resource

creydr:
# Changes

- :gift: Add helpers to create a pod (f5351b03e1c4e04224e0c5ffed9452fad0414bad)
- :broom: Move `WithAnnotations` and `WithLabels` to `pkg/manifest` as this is shared among different resources (91d02a331ca4e7e5180f7510cce7750451af9dc1)

Similar to #452

/kind enhancement


pkg/resources/job/options.go
```diff
@@ -29,22 +29,6 @@ func WithEnvs(envs map[string]string) manifest.CfgFn {
...
-	}
-}
-
-func WithLabels(labels map[string]string) manifest.CfgFn {
-	return func(cfg map[string]interface{}) {
-		if labels != nil {
-			cfg["labels"] = labels
-		}
-	}
-}
```
pierDipi, line 46: I'd leave those pointing to the new ones for discovery since if I type `job.` I get IDE suggestions and I don't need to remember "ah labels/annotations are in `manifest`"

pkg/resources/service/options.go
```diff
@@ -21,22 +21,6 @@ import (
...
-	}
-}
-
-func WithLabels(labels map[string]string) manifest.CfgFn {
-	return func(cfg map[string]interface{}) {
-		if labels != nil {
-			cfg["labels"] = labels
-		}
-	}
-}
```
pierDipi, line 38: Same here

pkg/resources/job/options.go
creydr: Yeah. But then all of them implement the same logic :/ WDYT about keeping it in the resource packages and referencing the one from manifest, like:

in job.go for example
```
var WithAnnotations = manifest.WithAnnotation
var WithLabels = manifest.WithLabels
```

pierDipi: Yes, that's what I meant with "pointing to the new ones"

creydr: :+1: Updated the PR

creydr:
/hold
need to add the `WithPort` option

creydr:
@pierDipi could you recheck on this PR?

pierDipi (approved):
/lgtm
/approve

PR #465, opened 2023-02-13
Add file-based image resolver

pierDipi:
Downstream we build images before running the tests and we don't use `ko`, so a file-based image resolver helps us referencing images differently.

### DIFF
--- a/pkg/images/file/file.go
+++ b/pkg/images/file/file.go
@@ -33,7 +33,7 @@ import (
 // knative.dev/reconciler-test/cmd/eventshub: quay.io/myregistry/eventshub
 // # ... other images ...
 func ImageProducer(filepath string) func(ctx context.Context, pack string) (string, error) {
-	images := make(map[string]string, 2 /* */)
+	images := make(map[string]string, 2)
 
 	bytes, err := os.ReadFile(filepath)
 	if err != nil {
--- a/pkg/images/file/file.go
+++ b/pkg/images/file/file.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2023 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package file
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"gopkg.in/yaml.v3"
+)
+
+// ImageProducer is a file-based image producer.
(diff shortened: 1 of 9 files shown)


pkg/images/file/file.go
```diff
@@ -0,0 +1,54 @@
...
+// ImageProducer is a file-based image producer.
+// The expected file format is a YAML file like the following:
+// <go-package-ref>: <image-ref>
+//
+// For example:
+//
+// knative.dev/reconciler-test/cmd/eventshub: quay.io/myregistry/eventshub
+// # ... other images ...
+func ImageProducer(filepath string) func(ctx context.Context, pack string) (string, error) {
+	images := make(map[string]string, 2 /* */)
```
creydr, line 36:
```suggestion
	images := make(map[string]string, 2)
```
Any reasons, why size/hint of 2?

pkg/images/file/file.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	if err := yaml.Unmarshal(bytes, images); err != nil {
+		panic("failed to unmarshal images file: " + err.Error() + "\nContent:\n" + string(bytes))
+	}
+
+	return func(ctx context.Context, pack string) (string, error) {
+		im, ok := images[pack]
+		if !ok {
+			return "", fmt.Errorf("image for package %s not found in %+v", pack, images)
```
creydr, line 50: Do we want to return an error? This means we would have to "override" all images IIUC. It would be great if it would be allowed to override only certain images, so e.g. some `ko://...` images could be still in place :shrug:

pkg/images/file/file_test.go
```diff
@@ -0,0 +1,54 @@
...
+		value string
+	}{
+		{
+			key:   "knative.dev/reconciler-test/cmd/eventshub",
+			value: "quay.io/myregistry/eventshub",
+		},
+		{
+			key:   "knative.dev/reconciler-test/cmd/eventshub2",
+			value: "quay.io/myregistry/eventshub",
+		},
```
creydr, line 41: If we want to return the error `image for package %s not found in %+v`, could we add a test here too?

pkg/images/file/file.go
pierDipi: Wouldn't that produce an empty image and so an error when deploying the resource?

pierDipi: or are you thinking that instead of having one image producer `xor` the other, we would need to compose image producers

creydr: > or are you thinking that instead of having one image producer xor the other, we would need to compose image producers?

more something like this. So in case it is not found in the defined image producer, it falls back to the default. But not sure if this is kind of unexpected/intransparent for the user... But it would allow to not have to override all used images

pierDipi: Yeah, that's why better to be explicit but also I don't have the use case of using both pre-built images and ko, so I'd leave that fallback behavior for now

pkg/images/file/file_test.go
pierDipi: Sure

pierDipi: Done!

creydr (approved):
/lgtm

PR #466, opened 2023-02-13
Allow testing with istio enabled

pierDipi:
This will allow start testing eventing-istio by using the existing
eventing test suite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -509,6 +509,16 @@ Run all instances of Noop feature, including instances in a Feature Set
 go test -v -count=1 -tags=e2e ./test/... --feature=Noop
 ```
 
+### Enable Istio sidecar injection
+
+Istio requires annotations on namespaces or pods to inject its sidecar, to enable
+Istio injection for reconciler-test resources, you can add the `istio.enabled`
+command line flag:
+
+```shell
+go test -v -count=1 -tags=e2e ./test/... --istio.enabled=true
+```
+
 #### Using Private Registries
 
 If a private registry is to be used with authentication, a Secret needs to be created in the `default` namespace called `kn-test-image-pull-secret` with the credentials.
--- a/README.md
+++ b/README.md
@@ -511,7 +511,7 @@ go test -v -count=1 -tags=e2e ./test/... --feature=Noop
 
 ### Enable Istio sidecar injection
 
-Istio requires annotations on namespaces or pods to inject its sidecar, to enable
+Istio requires annotations on pods and/or labels on namespaces to inject its sidecar, to enable
 Istio injection for reconciler-test resources, you can add the `istio.enabled`
 command line flag:
 
--- a/pkg/environment/istio.go
+++ b/pkg/environment/istio.go
@@ -1,0 +1,65 @@
+/*
+Copyright 2023 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package environment
+
+import (
+	"context"
+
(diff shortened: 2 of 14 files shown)


README.md
```diff
@@ -509,6 +509,16 @@ Run all instances of Noop feature, including instances in a Feature Set
 go test -v -count=1 -tags=e2e ./test/... --feature=Noop
 ```
 
+### Enable Istio sidecar injection
+
+Istio requires annotations on namespaces or pods to inject its sidecar, to enable
```
matzew, line 514: isn't is using labels?

pierDipi: both, labels on namespaces and annotations on pods, let me clarify

pkg/environment/istio.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+	}
+	return config.(*IstioConfig)
+}
+
+func withIstioNamespaceAnnotation(ns *corev1.Namespace) {
+	if ns.Annotations == nil {
+		ns.Annotations = make(map[string]string, 1)
+	}
+	ns.Annotations["istio-injection"] = "enabled"
```
matzew, line 50: > When you set the `istio-injection=enabled` label on a namespace and the injection webhook is enabled, any new pods that are created in that namespace will automatically have a sidecar added to them.

https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/

README.md
pierDipi, line 514:
```suggestion
Istio requires annotations on pods and/or labels on namespaces to inject its sidecar, to enable
```

pkg/environment/istio.go
pierDipi: Istio upstream is like that, downstream is different

README.md
pierDipi: Done

pkg/environment/istio.go
matzew: should it label the namespace (as well)?

matzew (approved):
/lgtm
/approve

/hold
not sure if we just want to annotate the namespace (given that the upstream doc speaks about labeling the NS). A bit confusing for me.

If we don't, feel free to remove the `hold`

pkg/environment/istio.go
pierDipi: Oh, I see what you mean

pierDipi:
@matzew updated, thanks!

pkg/manifest/options.go
```diff
@@ -43,10 +44,34 @@ func PodSecurityCfgFn(ctx context.Context, t feature.T) CfgFn {
...
+
+func appendToOriginal(original interface{}, additional map[string]interface{}) {
+	annotations := original.(map[string]interface{})
+	for k, v := range additional {
+		// Only add the unspecified ones
+		if _, ok := annotations[k]; !ok {
+			annotations[k] = v
 		}
 	}
+	return
```
creydr, line 74:
```suggestion
```

pkg/resources/deployment/options.go
```diff
@@ -58,14 +62,6 @@ func WithImagePullPolicy(ipp corev1.PullPolicy) manifest.CfgFn {
...
 }
 
-func WithPodAnnotations(annotations map[string]interface{}) manifest.CfgFn {
-	return func(cfg map[string]interface{}) {
-		if annotations != nil {
-			cfg["podannotations"] = annotations
-		}
-	}
-}
-
```
creydr, line 68: I think we could remove the unit test in deployment_test.go too (same applies for job & pod)

pierDipi: Wouldn't that reduce test coverage for `podannotations` ?

pkg/manifest/options.go
pierDipi: Automatic refactoring it's interesting sometimes

pkg/resources/deployment/options.go
creydr: Yes, but it would be the same as for the podSecurityContext, as the "implementation" is not done at this place. Or move the test to the `manifest` package? :shrug:

pierDipi: I don't have a resource yaml to test podannotations in manifest

creydr (approved):
/lgtm

pkg/resources/deployment/options.go
creydr: ack

PR #478, opened 2023-02-20
Allow verify service readiness when istio is installed

pierDipi:
The WaitForServiceReady was deploying a job without using the job package which means that the pod created out of the job is not annotated with istio injection annotation.

In addition, jobs with istio need to quit their sidecar when they ends so that the job can be considered "successful".


matzew (approved):
/lgtm
/approve

PR #479, opened 2023-02-20
Restart eventshub on failure

pierDipi:
When using eventshub on pods with istio sidecar injected, they terminate with failures since until the istio sidercar is ready network is not working so events will fail to get delivered to the configured destination.


matzew (approved):
/lgtm
/approve

PR #483, opened 2023-02-21
Improve pod ready condition check

pierDipi:
Pods in phase Running are not necessarly ready, so our old check is fragile and doesn't take into account the proper condition that ensure that a pod is actually ready.

For example, a pod in phase running might never progress to ready and with that we lose the debugging information we get by failing earlier like pod logs or pod conditions.


matzew (approved):
/lgtm
/approve

PR #485, opened 2023-02-23
Add teardown.on.fail flag

### LINKED ISSUE #486: Provide flag for skipping cleanup if test fails
It is sometimes necessary to investigate resources in the test namespace when the test fails because the test output might not have enough information. Like in the following case when the test fails and the only test output is following (which says the eventshub container failed, but the reason is unknown):
```
2023/02/20 11:36:04 {Pod test-qlgghnam sub4dlq-hdzqtgua  v1  }
    wait.go:382: Pod %s/%s is not running, dumping the pod object we got: test-qlgghnam sub4dlq-hdzqtgua
    wait.go:384: {
         "metadata": {
          "name": "sub4dlq-hdzqtgua",
          "namespace": "test-qlgghnam",
          "uid": "0f7a2a92-f1cf-4c20-be16-6d585792dd87",
          "resourceVersion": "222144",
          "creationTimestamp": "2023-02-20T16:36:04Z",
          "labels": {
           "app": "eventshub-sub4dlq-hdzqtgua"
          },
          "annotations": {
           "openshift.io/scc": "restricted-v2",
           "seccomp.security.alpha.kubernetes.io/pod": "runtime/default"
          },
          "managedFields": [
           {
            "manager": "rekt.test",
            "operation": "Update",
            "apiVersion": "v1",
            "time": "2023-02-20T16:36:04Z",
            "fieldsType": "FieldsV1",
            "fieldsV1": {
             "f:metadata": {
              "f:labels": {
               ".": {},
               "f:app": {}
              }
             },
             "f:spec": {
              "f:containers": {
               "k:{\"name\":\"eventshub\"}": {
                ".": {},
                "f:env": {
                 ".": {},
                 "k:{\"name\":\"EVENT_GENERATORS\"}": {
                  ".": {},
                  "f:name": {},
                  "f:value": {}
                 },
                 "k:{\"name\":\"EVENT_LOGS\"}": {
                  ".": {},
                  "f:name": {},
                  "f:value": {}
                 },
                 "k:{\"name\":\"K_CONFIG_LOGGING\"}": {
                  ".": {},
                  "f:name": {},
                  "f:value": {}
                 },
                 "k:{\"name\":\"K_CONFIG_TRACING\"}": {
                  ".": {},
                  "f:name": {},
                  "f:value": {}
                 },
                 "k:{\"name\":\"POD_NAME\"}": {
                  ".": {},
                  "f:name": {},
                  "f:valueFrom": {
                   ".": {},
                   "f:fieldRef": {}
                  }
                 },
                 "k:{\"name\":\"SYSTEM_NAMESPACE\"}": {
                  ".": {},
                  "f:name": {},
                  "f:valueFrom": {
                   ".": {},
                   "f:fieldRef": {}
                  }
                 }
                },
                "f:image": {},
                "f:imagePullPolicy": {},
                "f:name": {},
                "f:readinessProbe": {
                 ".": {},
                 "f:failureThreshold": {},
                 "f:httpGet": {
                  ".": {},
                  "f:path": {},
                  "f:port": {},
                  "f:scheme": {}
                 },
                 "f:periodSeconds": {},
                 "f:successThreshold": {},
                 "f:timeoutSeconds": {}
                },
                "f:resources": {},
                "f:securityContext": {
                 ".": {},
                 "f:allowPrivilegeEscalation": {},
                 "f:capabilities": {
                  ".": {},
                  "f:drop": {}
                 }
                },
                "f:terminationMessagePath": {},
                "f:terminationMessagePolicy": {}
               }
              },
              "f:dnsPolicy": {},
              "f:enableServiceLinks": {},
              "f:restartPolicy": {},
              "f:schedulerName": {},
              "f:securityContext": {
               ".": {},
               "f:runAsNonRoot": {},
               "f:seccompProfile": {
                ".": {},
                "f:type": {}
               }
              },
              "f:serviceAccount": {},
              "f:serviceAccountName": {},
              "f:terminationGracePeriodSeconds": {}
             }
            }
           }
          ]
         },
         "spec": {
          "volumes": [
           {
            "name": "kube-api-access-z6cbs",
            "projected": {
             "sources": [
              {
               "serviceAccountToken": {
                "expirationSeconds": 3607,
                "path": "token"
               }
              },
              {
               "configMap": {
                "name": "kube-root-ca.crt",
                "items": [
                 {
                  "key": "ca.crt",
                  "path": "ca.crt"
                 }
                ]
               }
              },
              {
               "downwardAPI": {
                "items": [
                 {
                  "path": "namespace",
                  "fieldRef": {
                   "apiVersion": "v1",
                   "fieldPath": "metadata.namespace"
                  }
                 }
                ]
               }
              },
              {
               "configMap": {
                "name": "openshift-service-ca.crt",
                "items": [
                 {
                  "key": "service-ca.crt",
                  "path": "service-ca.crt"
                 }
                ]
               }
              }
             ],
             "defaultMode": 420
            }
           }
          ],
          "containers": [
           {
            "name": "eventshub",
            "image": "registry.ci.openshift.org/openshift/knative-v1.6:knative-eventing-test-eventshub",
            "env": [
             {
              "name": "SYSTEM_NAMESPACE",
              "valueFrom": {
               "fieldRef": {
                "apiVersion": "v1",
                "fieldPath": "metadata.namespace"
               }
              }
             },
             {
              "name": "POD_NAME",
              "valueFrom": {
               "fieldRef": {
                "apiVersion": "v1",
                "fieldPath": "metadata.name"
               }
              }
             },
             {
              "name": "EVENT_LOGS",
              "value": "recorder,logger"
             },
             {
              "name": "EVENT_GENERATORS",
              "value": "receiver"
             },
             {
              "name": "K_CONFIG_LOGGING",
              "value": "{\"zap-logger-config\":\"{\\n  \\\"level\\\": \\\"info\\\",\\n  \\\"development\\\": false,\\n  \\\"outputPaths\\\": [\\\"stdout\\\"],\\n  \\\"errorOutputPaths\\\": [\\\"stderr\\\"],\\n  \\\"encoding\\\": \\\"json\\\",\\n  \\\"encoderConfig\\\": {\\n    \\\"timeKey\\\": \\\"ts\\\",\\n    \\\"levelKey\\\": \\\"level\\\",\\n    \\\"nameKey\\\": \\\"logger\\\",\\n    \\\"callerKey\\\": \\\"caller\\\",\\n    \\\"messageKey\\\": \\\"msg\\\",\\n    \\\"stacktraceKey\\\": \\\"stacktrace\\\",\\n    \\\"lineEnding\\\": \\\"\\\",\\n    \\\"levelEncoder\\\": \\\"\\\",\\n    \\\"timeEncoder\\\": \\\"iso8601\\\",\\n    \\\"durationEncoder\\\": \\\"\\\",\\n    \\\"callerEncoder\\\": \\\"\\\"\\n  }\\n}\\n\"}"
             },
             {
              "name": "K_CONFIG_TRACING",
              "value": "{\"backend\":\"zipkin\",\"debug\":\"true\",\"sample-rate\":\"1\",\"zipkin-endpoint\":\"[http://zipkin.istio-system.svc.cluster.local:9411/api/v2/spans\"}"](http://zipkin.istio-system.svc.cluster.local:9411/api/v2/spans/)
             }
            ],
            "resources": {},
            "volumeMounts": [
             {
              "name": "kube-api-access-z6cbs",
              "readOnly": true,
              "mountPath": "/var/run/secrets/kubernetes.io/serviceaccount"
             }
            ],
            "readinessProbe": {
             "httpGet": {
              "path": "/health/ready",
              "port": 8080,
              "scheme": "HTTP"
             },
             "timeoutSeconds": 1,
             "periodSeconds": 10,
             "successThreshold": 1,
             "failureThreshold": 3
            },
            "terminationMessagePath": "/dev/termination-log",
            "terminationMessagePolicy": "File",
            "imagePullPolicy": "IfNotPresent",
            "securityContext": {
             "capabilities": {
              "drop": [
               "ALL"
              ]
             },
             "runAsUser": 1001780000,
             "allowPrivilegeEscalation": false
            }
           }
          ],
          "restartPolicy": "Never",
          "terminationGracePeriodSeconds": 30,
          "dnsPolicy": "ClusterFirst",
          "serviceAccountName": "test-qlgghnam",
          "serviceAccount": "test-qlgghnam",
          "securityContext": {
           "seLinuxOptions": {
            "level": "s0:c42,c29"
           },
           "runAsNonRoot": true,
           "fsGroup": 1001780000,
           "seccompProfile": {
            "type": "RuntimeDefault"
           }
          },
          "imagePullSecrets": [
           {
            "name": "test-qlgghnam-dockercfg-7wprx"
           }
          ],
          "schedulerName": "default-scheduler",
          "tolerations": [
           {
            "key": "node.kubernetes.io/not-ready",
            "operator": "Exists",
            "effect": "NoExecute",
            "tolerationSeconds": 300
           },
           {
            "key": "node.kubernetes.io/unreachable",
            "operator": "Exists",
            "effect": "NoExecute",
            "tolerationSeconds": 300
           }
          ],
          "priority": 0,
          "enableServiceLinks": true,
          "preemptionPolicy": "PreemptLowerPriority"
         },
         "status": {
          "phase": "Pending",
          "qosClass": "BestEffort"
         }
        }
    wait.go:171: test-qlgghnam/sub4dlq-hdzqtgua is not Ready
        
        Resource: {
          "conditions": [
           {
            "type": "Initialized",
            "status": "True",
            "lastTransitionTime": "2023-02-20T16:36:04Z"
           },
           {
            "type": "Ready",
            "status": "False",
            "lastTransitionTime": "2023-02-20T16:36:04Z",
            "reason": "ContainersNotReady",
            "message": "containers with unready status: [eventshub]"
           },
           {
            "type": "ContainersReady",
            "status": "False",
            "lastTransitionTime": "2023-02-20T16:36:04Z",
            "reason": "ContainersNotReady",
            "message": "containers with unready status: [eventshub]"
           },
           {
            "type": "PodScheduled",
            "status": "True",
            "lastTransitionTime": "2023-02-20T16:36:04Z"
           }
          ]
         }
```

mgencur:
A similar flag is available in Serving. It's generally useful to keep the resources for inspection if the test fails because the test logs might not provide enough information.

# Changes

:gift:

- Introduce teardown.on.fail and disable it by default.
- Skip Teardown phase and namespace deletion if the test fails AND the flag is disabled
- If the test passes always execute the teardown phase and delete namespace

/kind enhancement

Fixes #486

**Release Note**

```release-note
The Teardown phases and namespace deletion are now skipped by default when a test fails. The new teardown.on.fail flag might be used to enforce teardown even if a test fails.
```

**Docs**

```docs
```

### DIFF
--- a/pkg/environment/flags.go
+++ b/pkg/environment/flags.go
@@ -30,8 +30,8 @@ var (
 	l = new(feature.Levels)
 	f = new(string)
 
-	ipFilePath    = new(string)
-	cleanupOnFail = new(bool)
+	ipFilePath     = new(string)
+	teardownOnFail = new(bool)
 )
 
 // InitFlags registers the requirement and state filter flags supported by the
@@ -60,7 +60,7 @@ func InitFlags(fs *flag.FlagSet) {
 
 	fs.StringVar(ipFilePath, "images.producer.file", "", "file path for file-based image producer")
 
-	fs.BoolVar(cleanupOnFail, "cleanup.on.fail", false, "Set this flag to do cleanup even if test fails.")
+	fs.BoolVar(teardownOnFail, "teardown.on.fail", false, "Set this flag to do teardown even if test fails.")
 }
 
 type stateValue struct {
--- a/pkg/environment/flags.go
+++ b/pkg/environment/flags.go
@@ -30,7 +30,8 @@ var (
 	l = new(feature.Levels)
 	f = new(string)
 
-	ipFilePath = new(string)
+	ipFilePath        = new(string)
+	skipCleanupOnFail = new(bool)
 )
 
 // InitFlags registers the requirement and state filter flags supported by the
@@ -58,6 +59,8 @@ func InitFlags(fs *flag.FlagSet) {
 	fs.StringVar(f, "feature", "", "run only Features matching `regexp`")
 
 	fs.StringVar(ipFilePath, "images.producer.file", "", "file path for file-based image producer")
+
+	fs.BoolVar(skipCleanupOnFail, "skip-cleanup-on-fail", true, "Set this flag to skip cleanup if test fails.")
 }
 
 type stateValue struct {
--- a/pkg/environment/flags.go
+++ b/pkg/environment/flags.go
@@ -60,7 +60,7 @@ func InitFlags(fs *flag.FlagSet) {
 
 	fs.StringVar(ipFilePath, "images.producer.file", "", "file path for file-based image producer")
 
-	fs.BoolVar(cleanupOnFail, "cleanup.on.fail", false, "Set this flag to cleanup even if test fails.")
+	fs.BoolVar(cleanupOnFail, "cleanup.on.fail", false, "Set this flag to do cleanup even if test fails.")
 }
 
 type stateValue struct {
--- a/pkg/environment/flags.go
+++ b/pkg/environment/flags.go
@@ -30,8 +30,8 @@ var (
 	l = new(feature.Levels)
 	f = new(string)
 
-	ipFilePath        = new(string)
-	skipCleanupOnFail = new(bool)
+	ipFilePath    = new(string)
+	cleanupOnFail = new(bool)
 )
 
 // InitFlags registers the requirement and state filter flags supported by the
@@ -60,7 +60,7 @@ func InitFlags(fs *flag.FlagSet) {
 
 	fs.StringVar(ipFilePath, "images.producer.file", "", "file path for file-based image producer")
 
-	fs.BoolVar(skipCleanupOnFail, "skip-cleanup-on-fail", true, "Set this flag to skip cleanup if test fails.")
+	fs.BoolVar(cleanupOnFail, "cleanup.on.fail", false, "Set this flag to cleanup even if test fails.")
 }
 
 type stateValue struct {
--- a/pkg/environment/magic.go
+++ b/pkg/environment/magic.go
@@ -45,7 +45,7 @@ func NewGlobalEnvironment(ctx context.Context, initializers ...func()) GlobalEnv
 		c:                initializeImageStores(ctx),
 		instanceID:       uuid.New().String(),
 		initializers:     initializers,
-		cleanupOnFail:    *cleanupOnFail,
+		teardownOnFail:   *teardownOnFail,
 	}
 }
 
@@ -60,7 +60,7 @@ type MagicGlobalEnvironment struct {
 	instanceID       string
 	initializers     []func()
 	initializersOnce sync.Once
-	cleanupOnFail    bool
+	teardownOnFail   bool
 }
 
 type MagicEnvironment struct {
@@ -84,7 +84,7 @@ type MagicEnvironment struct {
 	imagePullSecretName      string
 	imagePullSecretNamespace string
 
-	cleanupOnFail bool
+	teardownOnFail bool
 }
 
 const (
@@ -166,11 +166,11 @@ func (mr *MagicGlobalEnvironment) Environment(opts ...EnvOpts) (context.Context,
 	namespace := feature.MakeK8sNamePrefix(feature.AppendRandomString("test"))
 
 	env := &MagicEnvironment{
-		c:             mr.c,
-		l:             mr.RequirementLevel,
-		s:             mr.FeatureState,
-		featureMatch:  mr.FeatureMatch,
-		cleanupOnFail: mr.cleanupOnFail,
+		c:              mr.c,
+		l:              mr.RequirementLevel,
+		s:              mr.FeatureState,
+		featureMatch:   mr.FeatureMatch,
+		teardownOnFail: mr.teardownOnFail,
 
 		namespace:                namespace,
 		imagePullSecretName:      "kn-test-image-pull-secret",
@@ -330,11 +330,11 @@ func (mr *MagicEnvironment) Test(ctx context.Context, originalT *testing.T, f *f
 			// Special case for teardown timing
 			if timing == feature.Teardown {
 				if skip {
-					if mr.cleanupOnFail {
+					if mr.teardownOnFail {
 						// Prepend logging steps to the teardown phase when a previous timing failed.
 						steps = append(mr.loggingSteps(), steps...)
 					} else {
-						// When not doing cleanup only execute logging steps.
+						// When not doing teardown only execute logging steps.
 						steps = mr.loggingSteps()
 					}
 				}
--- a/pkg/environment/magic.go
+++ b/pkg/environment/magic.go
@@ -39,12 +39,13 @@ import (
 // expected to contain the configured Kube client already.
 func NewGlobalEnvironment(ctx context.Context, initializers ...func()) GlobalEnvironment {
 	return &MagicGlobalEnvironment{
-		RequirementLevel: *l,
-		FeatureState:     *s,
-		FeatureMatch:     regexp.MustCompile(*f),
-		c:                initializeImageStores(ctx),
-		instanceID:       uuid.New().String(),
-		initializers:     initializers,
+		RequirementLevel:  *l,
+		FeatureState:      *s,
+		FeatureMatch:      regexp.MustCompile(*f),
+		c:                 initializeImageStores(ctx),
+		instanceID:        uuid.New().String(),
+		initializers:      initializers,
+		skipCleanupOnFail: *skipCleanupOnFail,
 	}
 }
 
@@ -56,9 +57,10 @@ type MagicGlobalEnvironment struct {
 	c context.Context
 	// instanceID represents this instance of the GlobalEnvironment. It is used
 	// to link runs together from a single global environment.
-	instanceID       string
-	initializers     []func()
-	initializersOnce sync.Once
+	instanceID        string
+	initializers      []func()
+	initializersOnce  sync.Once
+	skipCleanupOnFail bool
 }
 
 type MagicEnvironment struct {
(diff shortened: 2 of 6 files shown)


pkg/environment/flags.go
```diff
@@ -58,6 +59,8 @@ func InitFlags(fs *flag.FlagSet) {
 	fs.StringVar(f, "feature", "", "run only Features matching `regexp`")
 
 	fs.StringVar(ipFilePath, "images.producer.file", "", "file path for file-based image producer")
+
+	fs.BoolVar(skipCleanupOnFail, "skip-cleanup-on-fail", true, "Set this flag to skip cleanup if test fails.")
```
cardil, line 63: Let's simplify this flag:

```suggestion
	fs.BoolVar(cleanupOnFail, "cleanup-on-fail", false, "Set this flag to cleanup even if test fails")
```

pkg/environment/magic.go
```diff
@@ -113,6 +117,9 @@ func (mr *MagicEnvironment) Finish() {
 	if mr.milestones != nil {
 		mr.milestones.Finished(result)
 	}
+	if result.Failed() && mr.skipCleanupOnFail {
+		return
+	}
 	if err := mr.DeleteNamespaceIfNeeded(); err != nil {
```
cardil, line 123: The `mr.DeleteNamespaceIfNeeded` seams like a better place to put this logic.

cardil (changes requested):
Thanks for doing this. Some minor questions.

pkg/environment/flags.go
pierDipi, line 63: The flag naming in this repository are dotted so for consistency we should follow that one, I'd call it

```
teardown.cleanup.onfail
```

pierDipi: and I would use teardown instead of cleanup since it's more aligned to the terminology of the timing

pkg/environment/magic.go
pierDipi: Why is it better?

pkg/environment/magic.go
```diff
@@ -327,10 +335,11 @@ func (mr *MagicEnvironment) Test(ctx context.Context, originalT *testing.T, f *f
 				// Prepend logging steps to the teardown phase when a previous timing failed.
 				if skip {
 					steps = append(mr.loggingSteps(), steps...)
+					// Teardown steps are executed only if test does not fail.
+					if !mr.skipCleanupOnFail {
+						skip = false
+					}
```
pierDipi, line 341: It would be easier to read and follow by putting this condition right after `if timing == feature.Teardown {` to avoid the already very deep nesting.

mgencur: It would't work because we want `steps = append(mr.loggingSteps(), steps...)` to be executed when test fails. If we put my condition at the beginning it won't append the steps :-/

pkg/environment/flags.go
mgencur: I'm using `cleanup.on.fail` . I think it's not required to put too much context in the name itself.

pkg/environment/magic.go
mgencur: It would be something like this: https://gist.github.com/mgencur/70a33c462f54963ef28d6fe903d510c9
But I'm not sure it's much better...

pkg/environment/magic.go
cardil: Because it's called `DeleteNamespaceIfNeeded`. The most obvious place to have a delete logic. 

Putting logic outside adds a proceeding, and succeeding relation withing the code, which is a pattern which should be avoided if possible.

pkg/environment/flags.go
mgencur: > and I would use teardown instead of cleanup since it's more aligned to the terminology of the timing

The word "cleanup" is more general. It includes Teardown steps and deleting the namespace. I'd use a word that covers both. But it's not a big deal.

pkg/environment/magic.go
pierDipi: The steps appended there are skipped anyway, so we need to append steps that we execute

pierDipi: > Would this be better?
> 
> ```
>                         if timing == feature.Teardown {
> 				if skip {
> 					// Prepend logging steps to the teardown phase when a previous timing failed.
> 					steps = append(mr.loggingSteps(), steps...)
> 				}
> 				if skip && mr.cleanupOnFail {
> 					// Teardown steps are executed only if test does not fail.
> 					skip = false
> 				}
> 			}
> ```

Yes, this is what I mean

pkg/environment/flags.go
pierDipi: As I see it, deleting the namespace is a special kind of teardown step

pierDipi: but I don't have a strong opinion, just making sure we consider the other option for naming things consistenty

pkg/environment/magic.go
mgencur: Yeah. OK. But I think we might also want to execute the logging steps even if skipping cleanup? I've pushed a new commit. Please review.

pkg/environment/flags.go
mgencur: OK. Makes sense too. Renamed.

pkg/environment/magic.go
```diff
@@ -324,12 +329,15 @@ func (mr *MagicEnvironment) Test(ctx context.Context, originalT *testing.T, f *f
...
-				// Prepend logging steps to the teardown phase when a previous timing failed.
 				if skip {
-					steps = append(mr.loggingSteps(), steps...)
+					if mr.teardownOnFail {
+						// Prepend logging steps to the teardown phase when a previous timing failed.
+						steps = append(mr.loggingSteps(), steps...)
+					} else {
+						// When not doing teardown only execute logging steps.
+						steps = mr.loggingSteps()
+					}
```
pierDipi, line 339: This is even nicer!

pierDipi (approved):
/lgtm
/approve

mgencur:
@pierDipi @cardil Thanks a lot for the reviews.

PR #491, opened 2023-02-28
Allow specifying parallel tests for parallelizable features

pierDipi:
The below pseudo-code will run `f1` and `f2` sequentially:
```
Environment.Test(ctx, t, f1)
Environment.Test(ctx, t, f2)
```
However, some features can be beneficial to run them in parallel
with other parallel features since most of our tests are waiting
for something to become ready.

This PR adds:
```
Environment.ParallelTest(ctx, t, f1)
Environment.ParallelTestSet(ctx, t, fs1)
```

to specify that a feature or a feature set can be run
in parallel.

# Changes

- Allow specifying parallel tests for parallelizable features

/kind <kind>

Fixes #

**Release Note**

```release-note
Allow specifying parallel tests for parallelizable features
```

### DIFF
--- a/pkg/environment/interfaces.go
+++ b/pkg/environment/interfaces.go
@@ -45,8 +45,16 @@ type Environment interface {
 	// Test will execute the feature test using the given Context and T.
 	Test(ctx context.Context, t *testing.T, f *feature.Feature)
 
+	// ParallelTest will execute the feature test using the given Context and T in parallel with
+	// other parallel features.
+	ParallelTest(ctx context.Context, t *testing.T, f *feature.Feature)
+
 	// TestSet will execute the feature set using the given Context and T.
 	TestSet(ctx context.Context, t *testing.T, fs *feature.FeatureSet)
+
+	// ParallelTestSet will execute the feature set using the given Context and T with each feature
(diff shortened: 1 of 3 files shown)


creydr:
@pierDipi so this is mostly a "wrapper" for `t.Parallel()`? So: 
```
Environment.ParallelTest(ctx, t, f1)
```
would be the same as 
```
t.Parallel()
Environment.Test(ctx, t, f1)
```

pierDipi:
That is not the same:

```
t.Parallel() // this signal that the whole test can run in parallel with other top-level tests

// t1 and t2 don't run in parallel
t.Run("t1", func(t *testing.T) {

})
t.Run("t2", func(t *testing.T) {

})
```

and 

```
t.Parallel() // this signal that the whole test can run in parallel with other top-level tests

// t1 and t2 _run_ in parallel
t.Run("t1", func(t *testing.T) {
  t.Parallel()
})
t.Run("t2", func(t *testing.T) {
  t.Parallel()
})
```

pierDipi:
With this PR I'm targeting the second case

creydr (commented):
only one NIT. otherwise LGTM

PR #495, opened 2023-03-03
Allow using a prespecified namespace for running tests

pierDipi:
For testing istio with eventing, I'd like to specify a specific
namespace so that it can be pre configured with some istio-specific
configurations.


creydr (approved):
/lgtm

PR #496, opened 2023-03-06
Support cronjobs

pierDipi:
SinkBinding tests use CronJobs, this will replace the `job` package in eventing

# Changes

- Support cronjobs

### DIFF
--- a/pkg/resources/cronjob/cronjob.yaml
+++ b/pkg/resources/cronjob/cronjob.yaml
@@ -1,0 +1,86 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ .name }}
+  namespace: {{ .namespace }}
+  {{ if .annotations }}
+  annotations:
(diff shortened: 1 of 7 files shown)


pkg/resources/cronjob/cronjob.yaml
```diff
@@ -0,0 +1,86 @@
...
+    {{ end }}
+  {{ end }}
+  {{ if .labels }}
+  labels:
+    {{ range $key, $value := .labels }}
+    {{ $key }}: {{ $value }}
+    {{ end }}
+  {{ end }}
+spec:
+  schedule: "* * * * *"
```
creydr, line 19: Could we parameterize the schedule too?

pierDipi: Once we need it?

creydr: Ok. Can do this as a follow up when we need it.
/lgtm

PR #502, opened 2023-04-03
Support setting Job labels for cron jobs

pierDipi:
If the SinkBinding selection mode is set to inclusion, in eventing we need to set the `bindings.knative.dev/include: true` label on the Job to allow SinkBinding to make changes to the resulting Job.


creydr (approved):
/lgtm

PR #519, opened 2023-05-04
Remove resources when terminating the environment

pierDipi:
When calling env.Finish() the resources that have been created by such environment should be removed.

So far, we were relying on the namespace deletion to clean up resources, however, if a precreated namespace is used for multiple tests and multiple environments, the cluster is full of resources that are not needed anymore since those tests were successfully run.

### DIFF
--- a/pkg/feature/feature.go
+++ b/pkg/feature/feature.go
@@ -199,12 +199,25 @@ func (f *Feature) References() []corev1.ObjectReference {
 //
 // Expected to be used as a StepFn.
 func (f *Feature) DeleteResources(ctx context.Context, t T) {
+	refFailedDeletion, err := DeleteResources(ctx, t, f.References())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	f.refsMu.Lock()
+	defer f.refsMu.Unlock()
+
+	f.refs = refFailedDeletion
+}
+
+func DeleteResources(ctx context.Context, t T, refs []corev1.ObjectReference) ([]corev1.ObjectReference, error) {
 	dc := dynamicclient.Get(ctx)
-	for _, ref := range f.References() {
+
+	for _, ref := range refs {
 
 		gv, err := schema.ParseGroupVersion(ref.APIVersion)
 		if err != nil {
-			t.Fatalf("Could not parse GroupVersion for %+v", ref.APIVersion)
+			return nil, fmt.Errorf("could not parse GroupVersion for %+v", ref.APIVersion)
 		}
 
 		resource := apis.KindToResource(gv.WithKind(ref.Kind))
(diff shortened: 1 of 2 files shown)


pkg/feature/feature.go
```diff
@@ -200,12 +200,25 @@ func (f *Feature) References() []corev1.ObjectReference {
 //
 // Expected to be used as a StepFn.
 func (f *Feature) DeleteResources(ctx context.Context, t T) {
```
mgencur, line 202: This function still has the comment:
```
// It doesn't fail when a referenced resource couldn't be deleted.
// Use References to get the undeleted resources.
```
We're now changing the behavior. When some resource couldn't be deleted we're failing the test. Is this intentional?

A few lines below we still assign this but it will be probably unused because we now fail the test with `t.Fatal immediately`:
```
f.refs = refFailedDeletion
```

pierDipi: weren't we failing even without this patch?
```golang
	err := wait.Poll(time.Second, 4*time.Minute, func() (bool, error) { // ... }
	if err != nil {
		LogReferences(refFailedDeletion...)(ctx, t)
		t.Fatalf("failed to wait for resources to be deleted: %v", err)
	}
```

https://github.com/knative-sandbox/reconciler-test/blob/dc48d52d7e3cf5b7d2c8078b8907c5df2522c0c1/pkg/feature/feature.go#L229-L260

mgencur: Hmm. OK. So this condition was broken earlier. And the comment is not valid for a long time. So, now I wonder why `f.refs = refFailedDeletion` is still there (probably a bug).
Let's remove the invalid comment for now. Not sure what to do about the refs.

pkg/feature/feature.go
```diff
@@ -256,13 +269,10 @@ func (f *Feature) DeleteResources(ctx context.Context, t T) {
...
 		LogReferences(refFailedDeletion...)(ctx, t)
-		t.Fatalf("failed to wait for resources to be deleted: %v", err)
+		return nil, fmt.Errorf("failed to wait for resources to be deleted: %v", err)
 	}
 
-	f.refsMu.Lock()
-	defer f.refsMu.Unlock()
-
-	f.refs = refFailedDeletion
+	return refFailedDeletion, nil
```
maschmid, line 275: Can `refFailedDeletion` ever be non-nil here? (whenever `refFailedDeletion` is assigned, an error is returned anyway?)

mgencur: Yeah. We do this:
```
if err != nil {
				refFailedDeletion = append(refFailedDeletion, ref)
				return false, fmt.Errorf("failed to get resource %+v %s/%s: %w", resource, ref.Namespace, ref.Name, err)
			}
```

pierDipi: Yes, that's correct and an existing problem, I'll drop it

pierDipi:
@maschmid @mgencur I believe I've addressed your comments in the last commits

mgencur:
/lgtm

Thanks!

PR #520, opened 2023-05-05
Remove dependency on prow

pierDipi:
knative.dev/pkg/test/prow imports a ton of other dependencies which
are not really need and just cause problems updating.

The only usage is getLocalArtifactsDir() function which is small
and replaced with a custom function.


matzew (approved):
/lgtm
/approve

remove it with :fire:

PR #524, opened 2023-05-10
added option to activate https for the reconciler test receiver

gabo1208:
still not final implementation and working on the e2e test

# Changes

- 🎁 added option to activate https for the reconciler test receiver

/kind enhancement

Fixes [#6482](https://github.com/knative/eventing/issues/6842)

**Release Note**

```release-note
added option to activate https for the reconciler test receiver
```

**Docs**

```docs

```

### DIFF
--- a/pkg/eventshub/receiver/receiver.go
+++ b/pkg/eventshub/receiver/receiver.go
@@ -34,6 +34,11 @@ import (
 	"knative.dev/reconciler-test/pkg/eventshub/dropevents"
 )
 
+const (
+	CertFileName = "cert.pem"
+	KeyFileName  = "key.pem"
+)
+
 // Receiver is the entry point for sinking events into the event log.
 type Receiver struct {
 	// Name is the name of this Receiver, used to filter if multiple observers.
@@ -50,6 +55,7 @@ type Receiver struct {
 	skipResponseCode    int
 	skipResponseHeaders map[string]string
 	skipResponseBody    string
+	useHTTPS            bool
 }
 
 type envConfig struct {
@@ -90,6 +96,9 @@ type envConfig struct {
 
 	// If events should be dropped, specify additional HTTP Headers to return in response.
 	SkipResponseHeaders map[string]string `envconfig:"SKIP_RESPONSE_HEADERS" default:"" required:"false"`
+
+	// If events can be received via https
+	UseHTTPS bool `envconfig:"USE_HTTPS" default:"false" required:"false"`
 }
 
 func NewFromEnv(ctx context.Context, eventLogs *eventshub.EventLogs) *Receiver {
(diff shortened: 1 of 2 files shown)


gabo1208:
cc @gab-satchi @pierDipi one thing I do not like is that we depend on filenames to easily adapt tsl to go http handlers, but if that is ok for now then the e2e add some env options and this is good to go, wdyt?

pkg/eventshub/receiver/receiver.go
```diff
@@ -136,23 +145,32 @@ func NewFromEnv(ctx context.Context, eventLogs *eventshub.EventLogs) *Receiver {
...
 		skipResponseHeaders: env.SkipResponseHeaders,
+		useHTTPS:            env.UseHTTPS,
 	}
 }
 
 // Start will create the CloudEvents client and start listening for inbound
 // HTTP requests. This is a is a blocking call.
 func (o *Receiver) Start(ctx context.Context, handlerFuncs ...func(handler http.Handler) http.Handler) error {
 	var handler http.Handler = o
+	addr := ":443"
```
pierDipi, line 156:
```suggestion
	addr := ":8443"
```

pierDipi: 433 is a privileged port

pierDipi:
> cc @gab-satchi @pierDipi one thing I do not like is that we depend on filenames to easily adapt tsl to go http handlers, but if that is ok for now then the e2e add some env options and this is good to go, wdyt?

I think the important aspect is how this is exposed to test writer through the eventshub options more than how it is implemented in eventshub itself, I'd like to see utilities to make that config really easy

pkg/eventshub/receiver/receiver.go
pierDipi: We need to add ports to the service resource as well

pierDipi:
> > cc @gab-satchi @pierDipi one thing I do not like is that we depend on filenames to easily adapt tsl to go http handlers, but if that is ok for now then the e2e add some env options and this is good to go, wdyt?
> 
> I think the important aspect is how this is exposed to test writer through the eventshub options more than how it is implemented in eventshub itself, I'd like to see utilities to make that config really easy

To expand on this:
- what if we have the TLS certs provisioning part completely hidden and transparent to the test writer? For example, by (automatically) creating a cert-manger Certificate when a simple option is configured `eventshub.EnforceTLS(...)`?
- how do we make it easy to get the associated CA certs with such eventshub TLS receiver? For example, by having a single self-signed CA per namespace configured on the environment

Those are just examples and ideas to make writing TLS tests easy

pierDipi:
Superseded by https://github.com/knative-sandbox/reconciler-test/pull/527

PR #525, opened 2023-05-12
add support for sending events to HTTPS endpoints in Eventshub

Rahul-Kumar-prog:
Signed-off-by: Rahul Kumar rahulkumarrsde@gmail.com

Fixes https://github.com/knative/eventing/issues/6840
# Changes

- Add support for sending events to HTTPS endpoints

**Release Note**

```Add support for sending events to HTTPS endpoints

```

**Docs**

```docs

```


go.mod
```diff
@@ -88,6 +88,7 @@ require (
 	k8s.io/klog/v2 v2.80.2-0.20221028030830-9ae4992afb54 // indirect
 	k8s.io/kube-openapi v0.0.0-20220803162953-67bda5d908f1 // indirect
 	k8s.io/utils v0.0.0-20221108210102-8e77b1f39fe2 // indirect
+	knative.dev/eventing v0.37.0
```
pierDipi, line 91: Unfortunately, we cannot depend on eventing here since reconciler-test is a dependency for eventing, we would create a cycle, so we need to duplicate that was implemented in eventing.

Take a look at @Vishal-Chdhry's comment here: https://github.com/knative/eventing/issues/6840#issuecomment-1546609588

PR #535, opened 2023-05-24
Deprecate MatchEvent and be explicit about matching kind

pierDipi:
Currently, Eventshub matcher library exposes a `MatchEvent` method,
such method is often used to match events but the intention is to
match events with kind `Received` or `Sent` as opposed to match
events of kind `Rejected`.

Technically, some tests might be matching Rejected events instead of
`Received` events (like "retry tests").

The new API exposes explicit methods to be intentional about what kind
of events to match with methods like: `MatchReceivedEvent`,
`MatchRejectedEvent`, `MatchSentEvent`, or `MatchResponseEvent`.


pkg/eventshub/assert/step.go
```diff
@@ -28,8 +28,40 @@ func (m MatchAssertionBuilder) Match(matchers ...eventshub.EventInfoMatcher) Mat
...
+func (m MatchAssertionBuilder) MatchResponseEvent(matchers ...cetest.EventMatcher) MatchAssertionBuilder {
+	m.matchers = append(m.matchers, MatchKind(eventshub.EventResponse))
+	m.matchers = append(m.matchers, MatchEvent(matchers...))
+	return m
+}
+
+// MatchEvent is a shortcut for Match(MatchEvent(), OneOf(MatchKind(eventshub.EventReceived), MatchKind(eventshub.EventSent)))
 func (m MatchAssertionBuilder) MatchEvent(matchers ...cetest.EventMatcher) MatchAssertionBuilder {
+	m.matchers = append(m.matchers, OneOf(
+		MatchKind(eventshub.EventReceived),
```
creydr, line 62: This probably will lead to some e2e test failures of existing tests...

creydr (approved):
/lgtm

PR #548, opened 2023-07-19
Support matching sender received peer certificates for certs rotation testing

pierDipi:
For certificate rotation testing, we need to be able to match that the eventshub sender while sending requests received the "latest" / "rotated" peer certificate.

The use would look like this:
```golang
	f.Assert("Sender received expected peer certificate", assert.OnStore(sourceName).
		MatchPeerCertificatesReceived(assert.MatchPeerCertificatesFromSecret(secretName, "tls.crt")).
		AtLeast(1),
	)
```

# Changes

- Support matching received peer certificates
- support defining matchers with context

Part of https://github.com/knative/eventing/issues/7039

### DIFF
--- a/pkg/eventshub/assert/step.go
+++ b/pkg/eventshub/assert/step.go
@@ -2,16 +2,21 @@ package assert
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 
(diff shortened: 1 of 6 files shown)


creydr (approved):
/lgtm

PR #549, opened 2023-07-19
Support rotating TLS certificates

pierDipi:
This PR adds a common StepFn for rotating a cert-manager certificates.

This PR in combination with https://github.com/knative-sandbox/reconciler-test/pull/548 will allow writing E2E cert-rotation tests that are generic for any implementation

Part of https://github.com/knative/eventing/issues/7039

### DIFF
--- a/resources/certificate/certificate.go
+++ b/resources/certificate/certificate.go
@@ -1,0 +1,199 @@
+/*
+Copyright 2023 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package certificate
+
+import (
+	"bytes"
+	"context"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"knative.dev/pkg/apis"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+	kubeclient "knative.dev/pkg/client/injection/kube/client"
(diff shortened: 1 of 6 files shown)


resources/certificate/certificate.go
```diff
@@ -0,0 +1,199 @@
...
+	certificateGVR = schema.GroupVersionResource{
+		Group:    "cert-manager.io",
+		Version:  "v1",
+		Resource: "certificates",
+	}
+)
+
+type RotateCertificate struct {
+	Certificate types.NamespacedName
+	Secret      types.NamespacedName
```
creydr, line 49: Couldn't we get the secret name from the certificate (`.spec.secretName` - which seems to be required)? Then the user would not have to look it up on its own, when calling `Rotate(RotateCertificate)`?

test/e2e/resources_test.go
```diff
@@ -59,3 +62,24 @@ func TestCronJobInstall(t *testing.T) {
...
+			certificate.RotateCertificate{
+				Certificate: types.NamespacedName{
+					Namespace: ns,
+					Name:      "test-certificate",
+				},
+				Secret: types.NamespacedName{
+					Namespace: ns,
+					Name:      "test-certificate-tls",
+				},
+			},
```
creydr, line 83: Could we add a check, that the certificate got really rotated? IIUC we only check that we don't run into errors here, or what am I missing?

pierDipi: It would be basically re-implement the functionality, it's basically a test of the test that you're asking?

pierDipi: Yes, that's correct, but to verify that it change and need to re-implement what I'm testing

pierDipi: See last commit

resources/certificate/certificate.go
pierDipi: Done

resources/certificate/certificate.go
```diff
@@ -0,0 +1,218 @@
...
+	certificateGVR = schema.GroupVersionResource{
+		Group:    "cert-manager.io",
+		Version:  "v1",
+		Resource: "certificates",
+	}
+)
+
+type RotateCertificate struct {
+	Certificate types.NamespacedName
+}
```
creydr, line 49:
```suggestion
type RotateCertificate types.NamespacedName
```
nit. so we could avoid the nesting.
```
foo := certificate.RotateCertificate{
	Certificate: types.NamespacedName{
		Namespace: "foo",
		Name:      "bar",
	},
}
```

But not blocking...

creydr (commented):
Nice!
/lgtm

Only one nit. Feel free to address it or skip it and remove the hold.
/hold

resources/certificate/certificate.go
pierDipi: Like things like "...Ctx" in https://github.com/knative-sandbox/reconciler-test/pull/548 could have been avoided with proper object/struct to work with instead of direct types

creydr: ack.
/unhold

PR #554, opened 2023-07-28
[release-1.10] Add name to port

pierDipi:
This helps with Istio testing since without name istio doesn't know the protocol 

- https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection
- https://istio.io/latest/docs/ops/common-problems/security-issues/#make-sure-you-are-not-using-http-only-fields-on-tcp-ports


ReToCode (approved):
Not a reviewer here, but /lgtm 😃

PR #557, opened 2023-07-28
Allow passing namespace for MatchPeerCertificatesFromSecret

pierDipi:
In Eventing, we need to use the system namespace rather than the test namespce.


Cali0707 (approved):
/lgtm

pierDipi:
Thank you!

PR #579, opened 2023-08-31
Copy pull secrets to ServiceAccount for eventshub

mgencur:
# Changes

- Same fix as [this commit](https://github.com/knative/eventing/commit/45e7a249a6a1e2c7f77ae7aa9588126912dd36a2) from eventing
- On top of them, copy the secrets to the newly created SA for each resource. So far we were only adding the pull secret to the default ServiceAccount in the namespace but for eventshub we create a new SA which needs the pull secrets as well.

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


pierDipi (approved):
/lgtm
/approve

PR #585, opened 2023-09-20
Option to override namespace for magic environment

mgencur:
This option is useful when it's necessary to test stuff across multiple namespaces which require some pre-setup steps. (e.g. testing traffic isolation with Istio between namespaces)

The test namespace can be defined for environment like this:
```
        global.Environment(
		environment.WithNamespace(namespace),
                ...
		environment.Managed(t),
	)
```

This option takes precedence over the --environment.namespace flag.

# Changes

-
-
-

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


pierDipi (approved):
/lgtm
/approve

PR #588, opened 2023-09-26
Add service account resource

creydr:
# Changes

- :gift: Add helpers to work with service accounts

/kind enhancement


pkg/resources/serviceaccount/options.go
```diff
@@ -0,0 +1,36 @@
...
+	WithAnnotations = manifest.WithAnnotations
+)
+
+func WithOwnerReferences(refs []metav1.OwnerReference) manifest.CfgFn {
+	return func(cfg map[string]interface{}) {
+		if refs != nil {
+			cfg["ownerReferences"] = refs
+		}
+	}
+}
```
creydr, line 36: we could also think about moving this to `manifest` (as done with `WithLabels` and `WithAnnotations`) as this might be helpful for other too. On the other hand: we are currently using it only here :shrug:

pierDipi: why do you need the owner references set?

pkg/resources/serviceaccount/serviceaccount.go
```diff
@@ -0,0 +1,76 @@
...
+	return func(ctx context.Context, t feature.T) {
+		if _, err := manifest.InstallYamlFS(ctx, yaml, cfg); err != nil {
+			t.Fatal(err)
+		}
+	}
+}
+
+// WaitForBeingAvailable waits until the given service account was created by
+// some 3rd party
+func WaitForBeingAvailable(name string, timing ...time.Duration) feature.StepFn {
```
pierDipi, line 54: This is to very that a given resource has created a service account?

pierDipi: Ideally, we don't enter in such implementation details, the contract and therefore the E2E behavior we need to verify is that "the subscriber will receive a JWT token emitted by the k8s configured issuer", the SA is the means we use for that but I think we should verify that the subscriber is receiving such tokens not that a SA is created (that's an implementation detail)

creydr: > This is to very that a given resource like "trigger" has created a service account?
Yes.

> Ideally, we don't enter in such implementation details in E2E tests
hmmm. I see your point, but how would you check then, that e.g. the trigger reconciler really creates the OIDC SA?

pkg/resources/serviceaccount/options.go
creydr: Wanted to make some tests with a set owner reference. But not really needed. I removed it...

pkg/resources/serviceaccount/serviceaccount.go
pierDipi: You would verify that the token you receive in the eventshub subscriber is what you expect (you need to do that anyway since that's actually tests everything E2E)

pierDipi: basically, we don't check "The SA is created" but "the token with the SA identity is received" and that's transitively testing that the SA exists

creydr:
As discussed, we will close this for now until we are clearer how to proceed for the eventhub

PR #591, opened 2023-09-27
:lady_beetle: Call the Istio' quit when the eventshub is closed

### LINKED ISSUE #590: The eventshub fails `pkg/k8s.WaitForPodReadyOrSucceededOrFail` on many Istio tests
We've seen a situation where the eventshub successfully exited before the `pkg/k8s.WaitForPodReadyOrSucceededOrFail` can be reported. In the case of testing with Istio, where every pod has its own side-container, this leads to timeout as the pod is in state:

```
containers with unready status: [eventshub]
```

See the log: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift-knative_eventing-istio/64/pull-ci-openshift-knative-eventing-istio-release-next-413-e2e-tests-aws-ocp-413/1706831632800944128

/kind bug

cardil:
# Changes

- :lady_beetle: Call the quit when the eventshub is closed

/kind bug

Fixes #590

**Release Note**

```release-note
The eventshub will call the Istio's /quitquitquit endpoint at exit
```

### DIFF
--- a/pkg/eventshub/eventshub.go
+++ b/pkg/eventshub/eventshub.go
@@ -23,6 +23,7 @@ import (
 	"golang.org/x/sync/errgroup"
 	"knative.dev/pkg/injection"
 	"knative.dev/pkg/logging"
+	"knative.dev/pkg/signals"
 )
 
 type envConfig struct {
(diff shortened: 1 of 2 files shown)


pkg/eventshub/eventshub.go
```diff
@@ -63,6 +63,7 @@ func Start(eventLogFactories map[string]EventLogFactory, eventGeneratorFactories
 	}
 
 	logging.FromContext(ctx).Info("Closing the eventshub process")
+	maybeQuitIstioProxy(ctx)
```
pierDipi, line 66: Can we use a deferred function? we need to call this even when panic happens or when we return early (above through Fatal)

cardil: Good point

cardil: Resolved in [`5fe3f46..4f44e5f`](https://github.com/knative-extensions/reconciler-test/compare/5fe3f46316380342fbaf0fe53507a1fff6eb8866..4f44e5f85295d121e491c0e7c6c6b871f7568ec3)

pierDipi (approved):
Thank you @cardil 

/lgtm
/approve

PR #595, opened 2023-10-06
Support OIDC in eventshub sender

creydr:
Fixes https://github.com/knative/eventing/issues/7333

# Changes

- :gift: Add OIDC support in eventshub sender. Adds the following options:
  - `OIDCValidToken()` adds a valid OIDC token to the request of the sender
  - `OIDCExpiredToken()` adds an OIDC token, which will expire in 10 minutes (allowed minimum) and adds a delay of 11 Minutes to sending to the request (via `InitialSenderDelay()`)
  - `OIDCInvalidAudience()` adds an OIDC token to the request but with an audience which differs from the sinks audience
  - `OIDCCorruptedSignature()` adds an OIDC token to the request with a corrupted tokens signature
  - `OIDCToken(token)` adds the given token to the request (allows to fully customize the used token)


creydr:
An alternative could be to let the eventhub get the audience (when the sender is setup via `StartSenderToResource`) and enable OIDC for the sender in general via an option (not via the "SinkAudience" option)

creydr:
Rebased

pierDipi (commented):
By having the sender creating the OIDC token, it will be hard to test what happens when sending an invalid token because it will be always valid, how do you plan to test failure cases?


IMHO, better to pass the token to the sender and have eventshub options that encapsulate the provisioning in order to:
- send invalid tokens for various reasons
  - expired
  - issued by a different issuer
  - for a different audience
- send valid token 

All of the failure or success cases could be options on eventshub sender and encapsulate the logic to provision the expected token

pierDipi:
The API for the raw sender could be similar to the SinkBinding contract, a directory from where to get the token, and then the eventshub options could configure _what_ value goes into that directory but on the test side

creydr:
> By having the sender creating the OIDC token, it will be hard to test what happens when sending an invalid token because it will be always valid, how do you plan to test failure cases?
> 
> IMHO, better to pass the token to the sender and have eventshub options that encapsulate the provisioning in order to:
> 
> * send invalid tokens for various reasons
>   
>   * expired
>   * issued by a different issuer
>   * for a different audience
>   * etc
> * send valid token
> 
> All of the failure or success cases could be options on eventshub sender and encapsulate the logic to provision the expected token valid or invalid token

I didn't want to do the token generation in the test/feature. For that I added for example the `SinkAudience` eventshub.Option to let the user set a custom audience (e.g. in case they want to test for a different audience - see for example https://github.com/knative/eventing/pull/7336/commits/e247119934fb1a0a2389e8c22e5be79a01d68332). So we can add more of those options to "customize" the token (even to specify a raw token).

Alternatively we could also add some kind of "TokenBuilder", which then gets passed to the eventshub sender as an option, which could be easier to extend :shrug:

pierDipi:
How do we test the other cases like expired tokens?

creydr:
> How do we test the other cases like expired tokens?

We could add another option e.g. to specify the expiry of a token, which then gets taken into account in `getOIDCToken()`

creydr:
Let me test something...

pierDipi:
That has the problem that every option needs to be built into the sender, I think sender should offer low level APIs, higher level logic can be offered as functions/options.

That also helps with test debugging, errors in the sender are not visible in the test output/logs

creydr:
@pierDipi: I did some tests with passing the token directly as an eventshub option. 
On the testing side, this leads to some extended StepFn definitions. E.g. to set the correct audience:

```
	// Send event with correct audience
	f.Requirement("install source", func(ctx context.Context, t feature.T) {
		address, err := k8s.Address(ctx, broker.GVR(), brokerName)
		if err != nil {
			t.Fatal(err)
		}

		token, err := auth.NewOIDCTokenProvider(ctx).GetJWT(types.NamespacedName{Namespace: namespace, Name: source1ServiceAccount}, *address.Audience)
		if err != nil {
			t.Fatalf("could not get token for source 1 service account: %w", err)
		}

		eventshub.Install(
			source2,
			eventshub.StartSenderToResource(broker.GVR(), brokerName),
			eventshub.EnableOIDCAuth,
			eventshub.OIDCToken(token),
			eventshub.InputEvent(eventForCorrectAudience),
		)(ctx, t)
	})
```

WDYT, about having an additional eventhub option, which allows to pass a token generator callback, which gets the context, the Audience of the Sink and the namespace of the source passed, so that we could do something like following:
```
        f.Requirement("install source", eventshub.Install(
		source2,
		eventshub.StartSenderToResource(broker.GVR(), brokerName),
		eventshub.EnableOIDCAuth,
		eventshub.OIDCTokenGenerator(func(ctx context.Context, sinkAudience, sourceNamespace string) (string, error) {
			return auth.NewOIDCTokenProvider(ctx).GetJWT(types.NamespacedName{
				Namespace: sourceNamespace, 
				Name: source1ServiceAccountName,
			}, sinkAudience)
		}),
		eventshub.InputEvent(eventForCorrectAudience),
	))
```

creydr:
I updated it to let the user specify the token via the `OIDCToken()` option.

Can be used like in the following:

```
// Send event with correct audience
	f.Requirement("install source", func(ctx context.Context, t feature.T) {
		address, err := k8s.Address(ctx, broker.GVR(), brokerName)
		if err != nil {
			t.Fatal(err)
		}

		token, err := auth.NewOIDCTokenProvider(ctx).GetJWT(types.NamespacedName{Namespace: namespace, Name: source1}, *address.Audience)
		if err != nil {
			t.Fatalf("could not get token for source 1 service account: %w", err)
		}

		eventshub.Install(
			source2,
			eventshub.StartSenderToResource(broker.GVR(), brokerName),
			eventshub.OIDCToken(token),
			eventshub.InputEvent(eventForCorrectAudience),
		)(ctx, t)
	})
```

IMO it would be better, if we have an **additional** eventhub option, which allows to pass a token generator callback, which gets the context, the Audience of the Sink and the namespace of the source passed, so that we could do something like following:
```
        f.Requirement("install source", eventshub.Install(
		source2,
		eventshub.StartSenderToResource(broker.GVR(), brokerName),
		eventshub.EnableOIDCAuth,
		eventshub.OIDCTokenGenerator(func(ctx context.Context, sinkAudience, sourceNamespace string) (string, error) {
			return auth.NewOIDCTokenProvider(ctx).GetJWT(types.NamespacedName{
				Namespace: sourceNamespace, 
				Name: source1ServiceAccountName,
			}, sinkAudience)
		}),
		eventshub.InputEvent(eventForCorrectAudience),
	))
```
But not sure, how this can be done with the current EventhubOption signature :thinking:

creydr:
@pierDipi can you recheck?

pkg/eventshub/resources.go
```diff
@@ -39,6 +41,13 @@ import (
 	"knative.dev/reconciler-test/pkg/resources/knativeservice"
 	"knative.dev/reconciler-test/pkg/resources/secret"
 	"knative.dev/reconciler-test/pkg/resources/service"
+	"knative.dev/reconciler-test/pkg/resources/serviceaccount"
+
+	authv1 "k8s.io/api/authentication/v1"
+)
+
+const (
+	OIDCTokenExiryMinutes = 10
```
pierDipi, line 50:
```suggestion
	OIDCTokenExpiryMinutes = 10
```

pkg/eventshub/resources.go
```diff
@@ -235,3 +273,65 @@ func getCACertsFromSecret(ctx context.Context) ([]byte, error) {
...
+		Spec: authv1.TokenRequestSpec{
+			Audiences: []string{
+				envs[OIDCSinkAudienceEnv],
+			},
+		},
+	}
+
+	if wantExpiredToken {
+		tokenRequest.Spec.ExpirationSeconds = pointer.Int64(OIDCTokenExiryMinutes * 60)
+	}
```
pierDipi: I'm not sure about the wantExpiredToken logic, can you describe how this provides an expired token?

pierDipi: I miss a "waiting for it to expire" somewhere

pierDipi:
Are you planning to have e2e tests for this feature in https://github.com/knative-extensions/reconciler-test/tree/main/test/e2e/eventshub ?

pkg/eventshub/resources.go
creydr: This part will create a token, which will expire in 10 minutes (minimal allowed value). At the `OIDCExpiredToken()`, we also delay the sending of the initial event by 10 +1 minutes: 
https://github.com/knative-extensions/reconciler-test/blob/4b20245817121808f93fa3895b75fe138b0ae14b/pkg/eventshub/options.go#L283-L285

That's also the reason, why I added `WaitForCompleted()` function, which returns, when the pod succeeded (what is the state of a sender after sending is done), so we can be sure the EventMatcher `Not()` only succeeds when the event was really not sent.

creydr:
> Are you planning to have e2e tests for this feature in https://github.com/knative-extensions/reconciler-test/tree/main/test/e2e/eventshub ?

Good point. We should add it, as soon as we have OIDC support in the receiver too (I'll create an issue for tracking it)

pkg/resources/serviceaccount/serviceaccount.go
```diff
@@ -0,0 +1,78 @@
...
+				return false, nil
+			}
+
+			return true, nil
+		})
+		if err != nil {
+			t.Fatalf("failed to get service account %s/%s: %v. Last error: %v", name, env.Namespace(), err, lastErr)
+		}
+	}
+}
```
pierDipi, line 78: Do we need this function?

pkg/resources/pod/pod.go
```diff
@@ -59,3 +67,31 @@ func registerImage(ctx context.Context, image string) error {
...
+
+			return pod.Status.Phase == corev1.PodSucceeded, nil
+		})
+
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+}
```
pierDipi, line 97: Can't we use https://github.com/knative-extensions/reconciler-test/blob/ffaf4f6f2d0ed37611fa7b9e57d893e64885fe02/pkg/k8s/wait.go#L74-L76 for this or extend it to be valid for any resource?

pkg/eventshub/resources.go
pierDipi: I see, we're configuring the sender delay

pkg/resources/serviceaccount/serviceaccount.go
creydr: :facepalm: no. sorry. Removed it

pkg/eventshub/resources.go
creydr: yes. I tried to state some addition information in the comment of `OIDCExpiredToken()`:
https://github.com/knative-extensions/reconciler-test/blob/4b20245817121808f93fa3895b75fe138b0ae14b/pkg/eventshub/options.go#L275-L283

pkg/eventshub/options.go
```diff
@@ -268,6 +268,45 @@ func InputMethod(method string) EventsHubOption {
...
+	return compose(envOption(OIDCGenerateValidTokenEnv, "true"), envOIDCEnabled())
+}
+
+// OIDCExpiredToken adds an expired OIDC token to the request. As the minimal
+// expiry for JWTs from Kubernetes are 10 minutes, the sender will delay the
+// send by 10 + 1 minutes.
+// This should be used in combination of an increase of the PollTimings (via
+// environment.WithPollTimings()) and the stepfunction
+// eventshub.WaitForCompleted() to make sure to wait long enough so the sender
+// was sending the expired token.
```
pierDipi, line 282: I'm not entire sure about this suggestion, I think, test side should just use `eventshub.OnStore()` utilities to make sure that the event was sent by the source as opposed to wait for something to be complete

pkg/resources/pod/pod.go
creydr: WaitForReadyOrDone would return as soon as the pod is ready. But in our case, we need it to be completed. Or what am I missing?

pierDipi: right, see my other comment, I'm not sure about this suggestion/pattern overall

pkg/eventshub/options.go
creydr: Updated it. I also added a new EnvOpts to configure the timings for the eventshub (`eventshub.WithRetryTimings()`)

pkg/eventshub/resources.go
```diff
@@ -235,3 +273,65 @@ func getCACertsFromSecret(ctx context.Context) ([]byte, error) {
...
+		parts := strings.Split(token, ".")
+		if len(parts) != 3 {
+			return "", fmt.Errorf("could not split token into header, payload and signature")
+		}
+
+		headerB64, payloadB64, signatureB64 := parts[0], parts[1], parts[2]
+
+		signature, err := base64.StdEncoding.DecodeString(signatureB64)
+		if err != nil {
+			// the signature may or may not encoded. When err != nil, it probably was not encoded --> use directly
```
pierDipi, line 327: Interesting, any reference to this?

From https://datatracker.ietf.org/doc/html/rfc7519 I read this

>  A JWT is represented as a sequence of URL-safe parts separated by
   period ('.') characters.  Each part contains a base64url-encoded
   value.

pierDipi: I think the `HMAC SHA-256` algorithm used to produce the signature might produce invalid characters for HTTP headers if that's the case

pkg/eventshub/event_info_store.go
```diff
@@ -30,16 +30,44 @@ import (
...
+
+// WithRetryTimings is an environment option to override default retry timings.
+func WithRetryTimings(interval, timeout time.Duration) environment.EnvOpts {
+	return func(ctx context.Context, env environment.Environment) (context.Context, error) {
+		return context.WithValue(ctx, timingsKey{}, timingsType{
+			interval: interval,
+			timeout:  timeout,
+		}), nil
+	}
+}
```
pierDipi, line 69: why not using `ContextWithPollTimings` and `PollTimingsFromContext` ?

creydr: We can use `environment.PollTimingsFromContext`. I just thought that `environment.PollTimingsFromContext` is used only for k8s pollings and not directly related to event polling in context of eventshub. I don't have a strong opinion in it...

pierDipi: I think it simpler to use with a single timing entrypoint, I understand the potential need to have more fine grained configuration but I'm not sure it's worth the complexity of understand "I have configured this timeout but the test is not working as expected" because it's not the correct timeout

creydr: ack. Will update it

creydr: updated

pkg/eventshub/event_info_store.go
```diff
@@ -30,14 +30,15 @@ import (
...
 	"knative.dev/reconciler-test/pkg/k8s"
 )
 
 const (
 	// The interval and timeout used for checking events
-	retryInterval = 4 * time.Second
-	retryTimeout  = 4 * time.Minute
+	DefaultRetryInterval = 4 * time.Second
+	DefaultRetryTimeout  = 4 * time.Minute
 )
```
pierDipi, line 42: I guess we can remove these constants now?

creydr: updated

pkg/eventshub/resources.go
```diff
@@ -235,3 +273,87 @@ func getCACertsFromSecret(ctx context.Context) ([]byte, error) {
...
+}
+
+func encodeBase64URL(src []byte) string {
+	s := base64.StdEncoding.EncodeToString(src)
+	s = strings.ReplaceAll(s, "+", "-")
+	s = strings.ReplaceAll(s, "/", "_")
+	s = strings.ReplaceAll(s, "=", "")
+
+	return s
+}
```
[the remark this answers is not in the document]
pierDipi:
```
base64.NewDecoder(base64.URLEncoding, <string>) io.Reader
```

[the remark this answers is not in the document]
creydr: #603 I need a break... :facepalm: :dizzy_face:

PR #607, opened 2023-10-13
Enable OIDCAuth for eventshub sender when sink has an audience

creydr:
Currently the user has to indicate via the eventhub option `OIDCValidToken` that a valid token should be added to the senders request.
This PR changes it to be the default behavior, if the sink has an audience set.

### DIFF
--- a/pkg/eventshub/options.go
+++ b/pkg/eventshub/options.go
@@ -268,11 +268,6 @@ func InputMethod(method string) EventsHubOption {
 	return envOption("INPUT_METHOD", method)
 }
 
-// OIDCValidToken adds a valid OIDC token to use for OIDC authentication to the request.
-func OIDCValidToken() EventsHubOption {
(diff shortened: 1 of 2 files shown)


pkg/eventshub/options.go
```diff
@@ -268,11 +268,6 @@ func InputMethod(method string) EventsHubOption {
 	return envOption("INPUT_METHOD", method)
 }
 
-// OIDCValidToken adds a valid OIDC token to use for OIDC authentication to the request.
-func OIDCValidToken() EventsHubOption {
-	return compose(envOption(OIDCGenerateValidTokenEnv, "true"), envOIDCEnabled())
-}
```
pierDipi, line 274: It might be useful for targeting raw services ?

pierDipi: it should be still doable

pierDipi (approved):
/lgtm
/approve

PR #612, opened 2023-10-17
Support OIDC in eventshub receiver

creydr:
Fixes https://github.com/knative/eventing/issues/7334

# Changes

- :gift: Add OIDC support to eventshub receiver

As soon as an eventshub receiver was setup with an audience (via `OIDCReceiverAudience(aud)`), it will check verify the JWTs of incoming requests (and respond with a 401, if failed).

**Hints for reviewers:**
* Added the `OIDCSinkAudience(aud)` option too, to define the audience of the sink, which is helpful/needed, if the audience of the sink can't be determined by the source itself (e.g. when the sink does not provide the `.status.address.audience` field)
* We can add the OIDC service account name to the eventInfo and then later base some asserter on it, but I didn't see a real usecase for it at that time of the OIDC implementation
* We verify the tokens with the TokenReviewAPI to not add another dependency to some OIDC library (e.g. coreos/go-oidc) and safe the setup.

```release-note
Support OIDC in eventshub receiver
```

### DIFF
--- a/pkg/eventshub/options.go
+++ b/pkg/eventshub/options.go
@@ -202,6 +202,11 @@ func DropEventsResponseHeaders(headers map[string]string) EventsHubOption {
 	)
 }
 
+// OIDCReceiverAudience sets the expected audience for received OIDC tokens on the receiver side
+func OIDCReceiverAudience(aud string) EventsHubOption {
+	return compose(envOption(OIDCReceiverAudienceEnv, aud), envOIDCEnabled())
+}
+
 // --- Sender options
 
 // InitialSenderDelay defines how much the sender has to wait (in millisecond), when started, before start sending events.
@@ -281,6 +286,11 @@ func OIDCExpiredToken() EventsHubOption {
 // OIDCInvalidAudience creates an OIDC token with an invalid audience
 func OIDCInvalidAudience() EventsHubOption {
 	return compose(envOption(OIDCGenerateInvalidAudienceTokenEnv, "true"), envOIDCEnabled())
+}
+
+// OIDCSinkAudience sets the Audience of the Sink
+func OIDCSinkAudience(aud string) EventsHubOption {
+	return oidcSinkAudience(&aud)
 }
 
 func oidcSinkAudience(aud *string) EventsHubOption {
--- a/pkg/eventshub/receiver/receiver.go
+++ b/pkg/eventshub/receiver/receiver.go
@@ -218,16 +218,17 @@ func (o *Receiver) ServeHTTP(writer http.ResponseWriter, request *http.Request) 
 		return
 	}
 
+	var statusCode int
 	var rejectErr error
 	if o.EnforceTLS && !isTLS(request) {
 		rejectErr = fmt.Errorf("failed to enforce TLS connection for request %s", request.URL.String())
+		statusCode = http.StatusBadRequest
 	}
 
-	var rejectErrStatusCode int
 	if o.oidcAudience != "" {
 		if err := o.validateJWT(request); err != nil {
 			rejectErr = err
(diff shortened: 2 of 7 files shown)


pkg/eventshub/receiver/receiver.go
```diff
@@ -269,7 +289,13 @@ func (o *Receiver) ServeHTTP(writer http.ResponseWriter, request *http.Request)
 		for headerKey, headerValue := range o.skipResponseHeaders {
 			writer.Header().Set(headerKey, headerValue)
 		}
-		writer.WriteHeader(http.StatusBadRequest)
+
+		if rejectErrStatusCode > 0 {
+			writer.WriteHeader(rejectErrStatusCode)
+		} else {
+			writer.WriteHeader(http.StatusBadRequest)
+		}
```
pierDipi, line 297: In `EventInfo` there is `StatusCode` are we storing the status code we return there (before sending it)? so that on the test side we can have more visibility into what's happening (and we can also match on that status code in the tests to clarify "rejected but why ?" )

it seems a generally useful feature but more important the more status codes we can emit here

pierDipi (approved):
A minor comment

/lgtm
/approve

pkg/eventshub/options.go
```diff
@@ -202,6 +202,11 @@ func DropEventsResponseHeaders(headers map[string]string) EventsHubOption {
 	)
 }
 
+// OIDCReceiverAudience sets the expected audience for received OIDC tokens on the receiver side
+func OIDCReceiverAudience(aud string) EventsHubOption {
+	return compose(envOption(OIDCReceiverAudienceEnv, aud), envOIDCEnabled())
```
pierDipi, line 207: I didn't see any usage of the "ENABLED" env variable, it seemed fine to just check "non empty audience", perhaps this option can just fail/panic when passing an empty string as audience

creydr: I used it to set the `withOIDCAuth` config, so that we only install the ClusterRoleBinding to create a token review, when we need it. We could use the ENV var of the RECEIVER_AUDIENCE too, but the `withOIDCAuth` seemed a bit clearer for me here. :shrug:

pierDipi: I see, sounds good

pkg/eventshub/receiver/receiver.go
creydr: :+1: 
Added it

creydr:
Failing due to
```
panic: test timed out after 10m0s
```
As the expiredToken test takes at least 10 minutes, I increased the timeout in 0145c0b840bae978dc9ec4a595809a377808ff45

creydr:
@pierDipi can you recheck? I only increased the timeout in 0145c0b840bae978dc9ec4a595809a377808ff45

creydr:
/cc @pierDipi
Could you recheck? I needed to increase some timeouts for the expiredtoken test

test/e2e/eventshub/oidc_test.go
```diff
@@ -0,0 +1,152 @@
...
+		environment.Managed(t),
+		environment.WithPollTimings(4*time.Second, 12*time.Minute),
+		knative.WithKnativeNamespace("knative-reconciler-test"),
+		knative.WithLoggingConfig,
+		knative.WithTracingConfig,
+		k8s.WithEventListener,
+	)
+
+	env.Test(ctx, t, validToken())
+	env.TestSet(ctx, t, invalidTokens())
```
pierDipi, line 51: Maybe `Parallel*` versions would speed this up?

pierDipi (approved):
/lgtm
/approve

PR #616, opened 2023-10-19
Improve logging message when deleting resources

pierDipi:
- Log the resource
- Log the events associated with such resource


creydr (approved):
Thanks!

/lgtm

PR #624, opened 2023-10-31
Configure sender with CA certs when provided

pierDipi:
This will enable us to run tests with both OIDC and TLS enabled.


Cali0707 (approved):
/lgtm

PR #626, opened 2023-11-06
use context poll interval/timeout in DeleteResources

maschmid:
- :broom: Use context poll interval/timeout in DeleteResources
/kind cleanup

Deleting resources with finalizer also requires reconciliation, so waiting for a resource to be deleted is in principle no different to waiting for it to become Ready. In some cases it makes sense to increase the timeouts for waiting for deleted resources. Therefore, we'll apply the same poll interval/timeouts for waiting for deletion as set in ContextWithPollTimings, instead of the original hardcoded values.

Also moves the ContextWithPollTimings and PollTimingsFromContext to `state` package to break dependency cycle

**Release Note**

```release-note
feature.DeleteResources now uses poll interval/timeouts from Context for waiting for the resources to be deleted, instead of the original hardcoded values
```

### DIFF
--- a/pkg/environment/timings.go
+++ b/pkg/environment/timings.go
@@ -23,6 +23,8 @@ import (
 	"knative.dev/reconciler-test/pkg/state"
 )
 
+// this has been moved to state pkg to break cycle between environment and feature package,
+// keeping the consts here for backwards API compatibility
 const (
 	DefaultPollInterval = 3 * time.Second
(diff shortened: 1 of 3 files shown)


maschmid:
image

also apparently hit memory limit, let's split the execution into separate go test invocations in Makefile

cardil (approved):
/lgtm

PR #630, opened 2023-11-10
Flags for setting poll timeout and poll interval

mgencur:
# Changes

- Introduce flags `poll.timeout` and `poll.interval` which can modify the default timeout (4m) and interval (2s). This is important for slower environments or envs with additional resources (e.g. Istio) which slow down deployments.

/kind <kind>

Fixes #

**Release Note**

```release-note

```

**Docs**

```docs

```


mgencur:
This will need a fix.

pierDipi (approved):
/lgtm
/approve

PR #639, opened 2023-12-04
Add matcher for used OIDC user name

creydr:
Add a event matcher, to check the OIDC user name of the received event.

This adds two new functions:
* `MatchOIDCUser(username)` to directly check if the given username matches the OIDC username in EventInfo
* `MatchOIDCUserFromResource(gvr, resourceName)` to check, if the OIDC username from the given resource matches the OIDC username in the EventInfo

This allows e.g. testcases like the following:
```
...
f.Alpha("Broker").
	Must("subscriber receives event from correct sender identity", eventassert.OnStore(sink).MatchWithContext(eventassert.MatchOIDCUserFromResource(triggerresources.GVR(), triggerName)).Exact(1))
```

### DIFF
--- a/pkg/eventshub/assert/event_info_matchers.go
+++ b/pkg/eventshub/assert/event_info_matchers.go
@@ -114,6 +114,20 @@ func MatchStatusCode(statusCode int) eventshub.EventInfoMatcher {
 	}
 }
 
+// MatchOIDCUser matches the OIDC username used for the request
+func MatchOIDCUser(username string) eventshub.EventInfoMatcher {
+	return func(info eventshub.EventInfo) error {
+		if info.OIDCUserInfo == nil {
+			return fmt.Errorf("event OIDC username don't match. Expected: '%s', Actual: nil", username)
+		}
+		if info.OIDCUserInfo.Username != username {
+			return fmt.Errorf("event OIDC username don't match. Expected: '%s', Actual: %s", username, info.OIDCUserInfo.Username)
+		}
+
+		return nil
+	}
+}
+
 // MatchHeartBeatsImageMessage matches that the data field of the event, in the format of the heartbeats image, contains the following msg field
 func MatchHeartBeatsImageMessage(expectedMsg string) cetest.EventMatcher {
 	return cetest.AllOf(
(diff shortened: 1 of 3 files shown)


pkg/eventshub/event_info.go
```diff
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
+	v1 "k8s.io/api/authentication/v1"
```
Cali0707, line 24: @creydr could you move this import to fix the linter error?

creydr: :+1:

creydr:
/cc @pierDipi @Cali0707 
ready for review

pkg/eventshub/assert/event_info_matchers.go
```diff
@@ -114,6 +122,59 @@ func MatchStatusCode(statusCode int) eventshub.EventInfoMatcher {
 	}
 }
 
+// MatchOIDCUser matches the OIDC username used for the request
+func MatchOIDCUser(username string) eventshub.EventInfoMatcher {
+	return func(info eventshub.EventInfo) error {
+		if info.OIDCUserInfo == nil {
+			return fmt.Errorf("event OIDC username don't match. Expected: '%s', Actual: nil", username)
```
Cali0707, line 129: Maybe we could clarify the error message a bit here? To me this would seem like we are saynig that that OIDC user name was `nil`, but actually I think the error here is that the `OIDCUserInfo` is `nil`. So maybe something like `event OIDC usernames don't match: no OIDC user info in the event`?

pkg/eventshub/assert/event_info_matchers.go
```diff
@@ -114,6 +122,59 @@ func MatchStatusCode(statusCode int) eventshub.EventInfoMatcher {
...
+		}
+		if info.OIDCUserInfo.Username != username {
+			return fmt.Errorf("event OIDC username don't match. Expected: '%s', Actual: %s", username, info.OIDCUserInfo.Username)
```
Cali0707, line 132: nit:
```suggestion
			return fmt.Errorf("event OIDC usernames don't match. Expected: '%s', Actual: %s", username, info.OIDCUserInfo.Username)
```

Cali0707 (commented):
Just some small nits that I think will make it easier to debug failures when using this, otherwise LGTM

pkg/eventshub/assert/event_info_matchers.go
creydr: :+1:

pkg/eventshub/assert/event_info_matchers.go
creydr: updated :+1:

Cali0707 (approved):
/lgtm

PR #652, opened 2024-01-11
Remove kubectl dependency in hack/update-cert-manager.sh

pierDipi:
Similar to https://github.com/knative/eventing/pull/7572


creydr (approved):
/lgtm

PR #658, opened 2024-01-24
Create standard global environment with provided rest config

Cali0707:
# Changes

- Add new function which creates a standard global environment using the provided rest config rather than parsing a new config.

This is needed for https://github.com/knative/eventing/pull/7626


creydr (approved):
/lgtm

PR #664, opened 2024-02-01
make MatchPeerCertificatesFromSecret work with certificate chains

maschmid:
# Changes
- :broom: make MatchPeerCertificatesFromSecret work with certificate chains

/kind enhancement

When using intermediate CAs, the secret can contain more than one certificate (a certificate chain), so the comparison
```
for _, cert := range info.Connection.TLS.PemPeerCertificates {
    if cert == string(value) 
```
cannot work

Instead, we parse all the  certificates in the chain from the secret with `pem.Decode` , and we make sure that each of the certificates is present in the `info.Connection.TLS.PemPeerCertificates`

### DIFF
--- a/pkg/eventshub/assert/step.go
+++ b/pkg/eventshub/assert/step.go
@@ -1,8 +1,10 @@
 package assert
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
+	"encoding/pem"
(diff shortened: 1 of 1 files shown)


matzew (approved):
/lgtm
/approve

PR #674, opened 2024-02-15
[WIP] feat: automatically register logstream

Cali0707:
# Changes

- Automatically register logstream for each commonly known component in knative-eventing namespace


Cali0707:
/cc @pierDipi 

This is WIP, please leave feedback!

pkg/environment/magic.go
```diff
@@ -28,13 +32,21 @@ import (
...
 	"knative.dev/reconciler-test/pkg/feature"
 	"knative.dev/reconciler-test/pkg/milestone"
 	"knative.dev/reconciler-test/pkg/state"
 )
 
+var (
+	componentNamesByNamespace = map[string][]string{
+		"knative-eventing": []string{"eventing-controller", "eventing-webhook", "imc-controller", "imc-dispatcher", "mt-broker-controller", "mt-broker-filter", "mt-broker-ingress", "kafka-broker", "kafka-channel", "kafka-controller", "kafka-sink", "kafka-source", "kafka-webhook", "knative-kafka"},
+	}
+)
```
pierDipi, line 48: If we make the logstream integration a `milestone.Emitter` we could configure this list based on the component we're testing similar to tracing gatherer or the existing `LogEmitter` https://github.com/knative-extensions/reconciler-test?tab=readme-ov-file#inspecting-zipkin-traces-for-failed-tests

```golang
ctx, env := global.Environment(
   // ....
   // Register the logstream listener which will gather logs from system components.
   logging.WithLogStream(..., components, ...)
)
```

pkg/environment/magic.go
```diff
@@ -395,11 +407,40 @@ func (mr *MagicEnvironment) test(ctx context.Context, originalT *testing.T, f *f
 	skipTeardown := false
 
 	originalT.Run(f.Name, func(t *testing.T) {
-
 		if isParallel(ctx) {
 			t.Parallel()
 		}
 
+		componentNames, ok := componentNamesByNamespace[mr.Namespace()]
```
pierDipi, line 414: This will probably never return a non empty list as `mr.Namespace()` uses the test-specific namespace, there is a function for getting the "system namespace" `knative.KnativeNamespaceFromContext(ctx)`

pkg/environment/magic.go
```diff
@@ -395,11 +407,40 @@ func (mr *MagicEnvironment) test(ctx context.Context, originalT *testing.T, f *f
...
+		if !ok {
+			componentNames = []string{}
+		}
+
+		artifacts := os.Getenv("ARTIFACTS")
+		file, err := os.CreateTemp(artifacts, fmt.Sprintf("%s-*.log", f.Name))
```
pierDipi, line 420: I would add a directory `logs/<mr.Namespace()>/<f.Name>.log`

pkg/environment/magic.go
pierDipi: Similar to the tracing emitter, we could export logs only when tests fail

pkg/environment/magic.go
Cali0707: Ah, I was looking at the function but it causes an import cycle :( maybe we can add a similar function to environment package?

pierDipi: This combined with the emitter comment might solve the import cycle problem?

Cali0707:
Closing in favour of opening a new PR based on the new logging in ekb

PR #681, opened 2024-02-28
Forwarder: Write received event to request

creydr:
Currently we have issues in the forwarder as the request body is read twice (one time to get the message/cloudevent and a second time when the request/clone is sent again (`o.httpClient.Do(req)`)).

This PR addresses it and makes sure the request which is send from the forwarder has the received event in its body.

Hint: I was also checking to clone the request earlier, but `Request.Clone()` seems not really clone the body (https://github.com/golang/go/blob/33013e8ea821629858643f24c55805f5ddf316b5/src/net/http/request.go#L383)


pierDipi (approved):
/lgtm
/approve

PR #689, opened 2024-03-18
fix eventshub forwarder pod template annotations

maschmid:
- :bug: Fix eventshub forwarder template "podannotations" placement

/kind bug


creydr (approved):
Oh. Good catch :)

PR #692, opened 2024-03-18
eventshub forwarder, don't clone Host header

maschmid:
- :broom: eventshub forwarder, don't clone the Host header of the original request


pkg/eventshub/forwarder/forwarder.go
```diff
@@ -165,6 +165,8 @@ func (o *Forwarder) ServeHTTP(writer http.ResponseWriter, request *http.Request)
 	req := request.Clone(requestCtx)
 	// It is an error to set this field in an HTTP client request.
 	req.RequestURI = ""
+	// We don't want to use the original request Host header, so drop it from the clone
+	req.Host = ""
```
creydr, line 169: Didn't we have issues in the past, when we removed the kn-namespace header? But I only remember vaguely. @mgencur wasn't this related to https://github.com/knative-extensions/reconciler-test/pull/572 somehow.

creydr (approved):

PR #708, opened 2024-04-23
Bump default Ko version

pierDipi:
It's been reported here as a bug https://github.com/knative/eventing/pull/7799#issuecomment-2025228385


creydr (approved):
/lgtm

PR #709, opened 2024-04-24
Add timestamps to log lines

pierDipi:
This adds timestamps to each log line emitted by the framework as it helps with debugging and correlating events and system logs with the test's execution


Cali0707 (approved):
/lgtm

Cali0707:
@pierDipi looks like it timed out again, maybe there's an actual problem?

PR #713, opened 2024-05-03
Fix forwarder body buffering

pierDipi:
Currently, the forwarder is duplicating headers the tests fail when testing components within an Istio mesh.


creydr (approved):
/lgtm

Thanks for fixing

PR #723, opened 2024-05-15
Increase default QPS and Burst to avoid getting rate limited

pierDipi:
The test client used by every running test and we've seen many instances where we get rate-limited for no reason.


creydr (approved):
/lgtm

PR #737, opened 2024-06-14
Add StatusCode to Sent EventInfo

pierDipi:
That helps with matching source events and response status code


pierDipi:
hmm, it breaks some tests and assumptions, so let's not do this

PR #749, opened 2024-08-12
Allow to configure OIDC subject name for eventshub sender

creydr:
Adds an option to set the OIDC identity to use for the eventshub sender

### DIFF
--- a/pkg/eventshub/options.go
+++ b/pkg/eventshub/options.go
@@ -334,7 +334,7 @@ func OIDCCorruptedSignature() EventsHubOption {
 	return compose(envOption(OIDCGenerateCorruptedSignatureTokenEnv, "true"), envOIDCEnabled())
 }
 
-// OIDCSubject sets the name of the OIDC subject to use by the sender. Defaults to "oidc-<eventshub-name>"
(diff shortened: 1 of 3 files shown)


creydr:
```
Error: init failed to get project from boskos: failed to get a "gke-project" from boskos: resources not found
```
/test integration-tests

pkg/eventshub/options.go
```diff
@@ -334,6 +334,11 @@ func OIDCCorruptedSignature() EventsHubOption {
 	return compose(envOption(OIDCGenerateCorruptedSignatureTokenEnv, "true"), envOIDCEnabled())
 }
 
+// OIDCSubject sets the name of the OIDC subject to use by the sender. Defaults to "oidc-<eventshub-name>"
```
Cali0707, line 337: nit: from the code it seems like I don't need to set this value unless I want a specific SA name, but from the comment it seemed like maybe I did need to use this option if I want the OIDC subject to be set
```suggestion
// OIDCSubject sets the name of the OIDC subject to use by the sender. If this option is not set, it defaults to "oidc-<eventshub-name>"
```

creydr: :+1: updated

PR #750, opened 2024-08-12
feat: add support for verifying the event format in the eventshub receiver

Cali0707:
This is needed to test changes like https://github.com/knative/eventing/pull/8151

# Changes

- Add options for verifying the event format in the eventshub receiver
- Add code to the receiver to verify the event format when the right env var is set

<!--
In addition, categorize the changes you're making using the "/kind" Prow command, example:


creydr (commented):
/lgtm

PR #884, opened 2026-04-23
Improve eventshub install stability

dsimansk:
# Changes

- Improve eventshub install stability

/cc @creydr @maschmid

### DIFF
--- a/pkg/eventshub/options.go
+++ b/pkg/eventshub/options.go
@@ -86,13 +86,14 @@ func StartSenderToResource(gvr schema.GroupVersionResource, name string) EventsH
 	}
 }
 
-// StartSenderToNamespacedResource starts the sender in the eventshub pointing to the provided resource
+// StartSenderToNamespacedResource starts the sender in the eventshub pointing to the provided resource.
+// It polls until the resource becomes addressable rather than doing a single point-in-time lookup.
 // This can be used together with InputEvent, AddTracing, EnableIncrementalId, InputEncoding and InputHeader options
 func StartSenderToNamespacedResource(gvr schema.GroupVersionResource, name, namespace string) EventsHubOption {
 	return func(ctx context.Context, envs map[string]string) error {
-		u, err := k8s.NamespacedAddress(ctx, gvr, name, namespace)
+		u, err := k8s.WaitForAddress(ctx, gvr, name)
 		if err != nil {
-			return err
+			return fmt.Errorf("timed out waiting for resource %v named %s to become addressable: %w", gvr, name, err)
 		}
 		if u == nil {
 			return fmt.Errorf("resource %v named %s is not addressable", gvr, name)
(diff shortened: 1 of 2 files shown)


pkg/eventshub/resources.go
```diff
@@ -85,12 +84,11 @@ func Install(name string, options ...EventsHubOption) feature.StepFn {
 		}
 		env := environment.FromContext(ctx)
 		namespace := env.Namespace()
-		log := logging.FromContext(ctx)
 
 		// Compute the user provided envs
 		envs := make(map[string]string)
 		if err := compose(options...)(ctx, envs); err != nil {
-			log.Fatalf("Error while computing environment variables for eventshub: %s", err)
+			t.Fatalf("Error while computing environment variables for eventshub: %s", err)
```
dsimansk, line 91: log.Fatalf and transitively zap logger calls `os.Exit(1)` exiting immediately without any cleanup, debug collection or other test steps.

pkg/eventshub/options.go
```diff
@@ -86,13 +86,14 @@ func StartSenderToResource(gvr schema.GroupVersionResource, name string) EventsH
...
 }
 
-// StartSenderToNamespacedResource starts the sender in the eventshub pointing to the provided resource
+// StartSenderToNamespacedResource starts the sender in the eventshub pointing to the provided resource.
+// It polls until the resource becomes addressable rather than doing a single point-in-time lookup.
 // This can be used together with InputEvent, AddTracing, EnableIncrementalId, InputEncoding and InputHeader options
 func StartSenderToNamespacedResource(gvr schema.GroupVersionResource, name, namespace string) EventsHubOption {
 	return func(ctx context.Context, envs map[string]string) error {
-		u, err := k8s.NamespacedAddress(ctx, gvr, name, namespace)
+		u, err := k8s.WaitForAddress(ctx, gvr, name)
```
dsimansk, line 94: Specifically for e2e tests with Eventing + Istio, the address might not be populated yet.

pkg/eventshub/resources.go
maschmid: `t.Fatal` doesn't?

dsimansk:
Downstream test fail is unrelated, fixed in: https://github.com/knative-extensions/reconciler-test/pull/885

pkg/eventshub/options.go
maschmid: In which tests, specifically?

pkg/eventshub/resources.go
dsimansk: it should propagate the error back to calling `feature.T` to handle or to call teardown etc.

pkg/eventshub/options.go
maschmid: The tests generally wait for readiness before starting sender, so I am wondering which addersasble resource is Ready without having an address...  (I don't want this change to potentially hide actual issues)

dsimansk: https://github.com/knative-extensions/eventing-istio/pull/1007 running the applied patches here.

https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/knative-extensions_eventing-istio/1007/integration-tests_eventing-istio_main/2047028219114164224

In scenarios with latest Istio + Kafka + SASL auth.

maschmid: Do you have a run without the patches from which we can see that this problem occurred?

maschmid: Anyway, replacing NamespacedAddress with WaitForAddress looks wrong to me,  I expect that would break tests that explicitly use resources from different namespace (e.g. the CrossNamespaceEventLinks test)

dprotaso:
cycling to pick up latest github actions changes

dprotaso:
huh - dont' know if that worked haha

dprotaso:
do we need to bump k8s version?

pkg/eventshub/options.go
dsimansk: Indeed tests are intermittently failing or passing with or without the patch regarless:

Passed without patch: https://github.com/knative-extensions/eventing-istio/pull/1010
Failed without patch: https://github.com/knative-extensions/eventing-istio/pull/1008

dprotaso:
@dsimansk @creydr is this good to merge?

maschmid (changes requested):
replacing NamespacedAddress with WaitForAddress looks wrong to me, I expect that would break tests that explicitly use resources from different namespace (e.g. the CrossNamespaceEventLinks test)

Additionaly, I don't think the added wait should be needed. In general, the tests wait for the readiness before starting sender, so we might be hiding actual issues by adding an additional wait here.

dsimansk:
It's unneeded change for now. Testing it in eventing-istio CI run didn't prove that stability improved. 

/close

